PR #5, opened 2020-12-28
A hack to convert weird WL symbols to unicode

### LINKED ISSUE #4: Integrate[...] prints unicode F74C instead of "d" in "dx"
This is a follow up to https://github.com/mathics/Mathics/issues/206:

> I can reproduce this by opening http://www.mathics.net/ and doing:
> 
> ```
> Integrate[y[x], x]
> ```
> 
> which gives:
> 
> ```
>  ∫y[x]x
> ```

rocky:
For example "d" in "dx" is like this.

Fixes #4

### DIFF
--- a/mathics_django/web/views.py
+++ b/mathics_django/web/views.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 # -*- coding: utf-8 -*-
 
+import re
 import sys
 import traceback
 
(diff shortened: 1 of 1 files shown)


rocky:
Would like to get this into the next docker image since Combinatorica may be promoted by Steve Skeina. 

We can revert after this is in a more common location or addressed more properly.

PR #7, opened 2021-01-01
FullForm & OutputForm should not use MathML

rocky:
@mmatera @GarkGarcia 

This change is very important. Please look over carefully. 

(* Begin Rant *)
There has been some fundamental wrongness of thinking that I am trying to undo here. 
Decisions about how to format output is not a one-size-fits-all kind of thing and can be declared in advance of evaluation as a single value (like "xml" for Django). 

Instead, the format of result of an evaluation is based on: 
* the kind of result produced,
* the capabilities of the front end, and 
* any format desire specified, 

Specifically in the Django interface, not everything should be formatted as MathML. And if IWolfram does that too, that is also probably wrong. 

Also, I don't get why we treat XML as synomymous for MathML. They are not the same. Or that graphics for TeX is asymptote.

Formatting for particular backends like SVG, MathML and aysmptote need to come out of the core. 

That should be addressed down the line. 

(* End Rant *)

This PR starts to put final formatting decisions more in control of mathics-django.

To do this we set the output format to be "unformatted". This prevents Mathics Core from trying to make any decision and allows mathics-django's web/format.py to make decisions about the final formatting.

In particular, we want OutpuForm and FullForm *not* to produce MathML.
Instead it should produce plain-ol' text which you can cut and paste.

Worse, OutputForm[Rectangle[]] was producing a graphic when it shouldn't have been.

There still seems to be a bug with FullForm not expanding graphics primitives, but that can come later.

I should also mention something about matplotib graphics that has been working crudely, (but still working) in the Django interface. Check out the pymathics-graph project and in the README.rst there are some simple commands that you can try in either mathicsscript of mathics-django. 

In ths PR, that has gotten a little bit better because we render matplotlib filenames better now  since
that is no longer MathML, but cut and pateable text. It is still not clickable: a html link would be nice, but that exceeds my knowledge of how to get done in this code with django.

### DIFF
--- a/mathics_django/web/format.py
+++ b/mathics_django/web/format.py
@@ -9,7 +9,20 @@ import networkx as nx
 import tempfile
 
 
+FORM_TO_FORMAT = {
+    "System`MathMLForm": "xml",
+    "System`TeXForm": "tex",
+    "System`FullForm": "text",
+    "System`OutputForm": "text",
+}
+
 def format_output(obj, expr, format=None):
+    """
+    Handle unformatted output using the *specific* capabilities of mathics-django.
+
+    evaluation.py format_output() from which this was derived is similar but
+    it can't make use of a front-ends specific capabilities.
+    """
     if format is None:
         format = obj.format
 
@@ -18,22 +31,28 @@ def format_output(obj, expr, format=None):
 
     from mathics.core.expression import Expression, BoxError
 
+    # For some expressions, we want formatting to be different.
+    # In particular for FullForm output, we dont' want MathML, we want
+    # plain-ol' text so we can cut and paste that.
+
     expr_type = expr.get_head_name()
-    if expr_type == "System`MathMLForm":
-        format = "xml"
+    if expr_type in ("System`MathMLForm", "System`TeXForm"):
+        # For these forms, we strip off the outer "Form" part
+        format = FORM_TO_FORMAT[expr_type]
         leaves = expr.get_leaves()
         if len(leaves) == 1:
             expr = leaves[0]
-    elif expr_type == "System`TeXForm":
-        format = "tex"
-        leaves = expr.get_leaves()
-        if len(leaves) == 1:
-            expr = leaves[0]
+
+    if expr_type in ("System`FullForm", "System`OutputForm"):
+        result = Expression("StandardForm", expr).format(obj, expr_type)
+        return str(result)
     elif expr_type == "System`Graphics":
         result = Expression("StandardForm", expr).format(obj, "System`MathMLForm")
         ml_str = result.leaves[0].leaves[0]
         # FIXME: not quite right. Need to parse out strings
         # display_svg(str(ml_str))
+
+    # This part was derived from and the same as evaluation.py format_output.
 
     if format == "text":
         result = expr.format(obj, "System`OutputForm")
@@ -42,10 +61,11 @@ def format_output(obj, expr, format=None):
     elif format == "tex":
         result = Expression("StandardForm", expr).format(obj, "System`TeXForm")
     elif format == "unformatted":
+        # This part is custom to mathics-django:
         if str(expr) == "-Graph-" and hasattr(expr, "G"):
             return format_graph(expr.G)
         else:
-            result = expr.format(obj, "System`OutputForm")
+            result = Expression("StandardForm", expr).format(obj, "System`MathMLForm")
     else:
         raise ValueError
 
--- a/mathics_django/web/models.py
+++ b/mathics_django/web/models.py
@@ -9,6 +9,7 @@ from django.contrib.sessions.models import Session
 
 from mathics.core.definitions import Definitions
 from mathics.core.evaluation import Evaluation, Output
+from mathics_django.web.format import format_output
 
 
 class WebOutput(Output):
@@ -21,9 +22,14 @@ def get_session_evaluation(session):
     evaluation = _evaluations.get(session.session_key)
     if evaluation is None:
         definitions = Definitions(add_builtin=True)
+        # We set the formatter to "unformatted" so that we can use
+        # our own custom formatter that understand better how to format
+        # in the context of mathics-django.
+        # Previously, one specific format, like "xml" had to fit all.
         evaluation = Evaluation(
-            definitions, format='xml', output=WebOutput())
+            definitions, format='unformatted', output=WebOutput())
         _evaluations[session.session_key] = evaluation
+        evaluation.format_output = lambda expr, format: format_output(evaluation, expr, format)
     return evaluation
 
 
@@ -52,8 +58,8 @@ class Query(models.Model):
 
 
 class Worksheet(models.Model):
-    user = models.ForeignKey(User, 
-                             related_name='worksheets', 
+    user = models.ForeignKey(User,
+                             related_name='worksheets',
                              null=True, on_delete=models.CASCADE)
 
     name = models.CharField(max_length=30)
--- a/mathics_django/web/views.py
+++ b/mathics_django/web/views.py
@@ -155,14 +155,7 @@ def query(request):
                 results.append(Result(evaluation.out, None, None))  # syntax errors
                 evaluation.out = []
                 continue
-            # START HERE.
-            # Handle unformatted.
             result = evaluation.evaluate(expr, timeout=settings.TIMEOUT)
-            if str(result.last_eval) == "-Graph-":
-                from mathics_django.web.format import format_graph
-                svg_path = format_graph(result.last_eval.G)
-                result.result="""<math><mi href="file://%s">Graph file://%s</mi></math>""" % (svg_path, svg_path)
-
             if result.result is not None:
                 result.result = replace_wl_to_unicode(result.result)
             results.append(result)  # syntax errors


mmatera:
> @mmatera @GarkGarcia
> 
> This change is very important. Please look over carefully.
> 
> (* Begin Rant *)
> There has been some fundamental wrongness of thinking that I am trying to undo here.
> Decisions about how to format output is not a one-size-fits-all kind of thing and can be declared in advance of evaluation as a single value (like "xml" for Django).
> 
> Instead, the format of result of an evaluation is based on:
> 
>     * the kind of result produced,
> 
>     * the capabilities of the front end, and
> 
>     * any format desire specified,
> 
> 
> Specifically in the Django interface, not everything should be formatted as MathML. And if IWolfram does that too, that is also probably wrong.

Actually, in IWolfram I hit the same problem. As a user of WMA notebooks, I started thinking that the output of a formal expression should be (latex) mathematical formula. Then, it becomes evident that this kind of output avoids a very important user action: to copy the output, modify it a little bit, and use it again as a new input. The solution I found for that was to have "InputForm" form as the output by default and, by adding a `//MatrixForm` at the end of an expression, get a MathML output.
Apart from expressions, there are several other possible outputs. For example, graphics, sounds (from `Play`), 3D graphics, warnings, Print's, HTML, Javascript,  etc.  which can be produced by the WMA kernel and be handled by the frontend. Jupyter has a rich interface to handle all these kind of objects,  and usually you can send several possible outputs and let to the frontend to choose the best choice, according to the available interface (text / web / qt  interfaces, for example). The only problem is that you have to build all these different outputs.


> 
> Also, I don't get why we treat XML as synomymous for MathML. They are not the same. Or that graphics for TeX is asymptote.
>
The think is WMA has MathMLForm but not XMLForm to convert expressions. Indeed, the idea was to use XML to format different kind of outputs which can be sent to the Django interface. Then, you can send a SVG picture, a MathML formula, or any other object, and also you can nest formulas inside graphics, inside formulas. This is why when you started touching the format for graphics, text output inside MathML stopped working. What we need here is a more clear, robust, and consistent way to get the same behavior.
  
> Formatting for particular backends like SVG, MathML and aysmptote need to come out of the core.
> 
> That should be addressed down the line.
> 
> (* End Rant *)
> 
> This PR starts to put final formatting decisions more in control of mathics-django.
> 
> To do this we set the output format to be "unformatted". This prevents Mathics Core from trying to make any decision and allows mathics-django's web/format.py to make decisions about the final formatting.
> 
> In particular, we want OutpuForm and FullForm _not_ to produce MathML.

This has two layers: on the one hand, it would be nice to have the possibility of showing beautiful formulas, but also in a way easy to reuse outputs as new inputs.

WMA notebook allows this with machinery that we do not have available by now: It has a native interpreter that builds the formulas, in a way that if you copy a part of it, you get something evaluable. Something less complicated that I thought for IWolfram was to send a MathML expression as an output to the frontend, but with a "onclick" event that sends the "InputForm" expression to the clipboard, or to a new cell.

> Instead it should produce plain-ol' text which you can cut and paste.
> 
> Worse, OutputForm[Rectangle[]] was producing a graphic when it shouldn't have been.
> 
Graphics have another layer of complexity: on the one hand, they should be able to support formulas inside, for axes labels, captions, etc, ( which is something that asymptote or mpl could handle now, but not 10 years ago). Also, a visual representation of these graphics be built just if the interface support graphics, and in a format that can be used by the specific front-end. 
For all of this, I guess that the best would be to provide support to generate different kind of graphics formats (svg, png) from the mathics kernel, but to control the "makeboxes" for Graphics and Graphics3D on the different frontends.

> There still seems to be a bug with FullForm not expanding graphics primitives, but that can come later.
> 
> I should also mention something about matplotib graphics that has been working crudely, (but still working) in the Django interface. Check out the pymathics-graph project and in the README.rst there are some simple commands that you can try in either mathicsscript of mathics-django.
> 
> In ths PR, that has gotten a little bit better because we render matplotlib filenames better now since
> that is no longer MathML, but cut and pateable text. It is still not clickable: a html link would be nice, but that exceeds my knowledge of how to get done in this code with django.
OK.

Regarding the PR, I think it is OK at this stage at least.

rocky:
> > Specifically in the Django interface, not everything should be formatted as MathML. And if IWolfram does that too, that is also probably wrong.
>
> Actually, in IWolfram I hit the same problem. As a user of WMA notebooks, I started thinking that the output of a formal expression should be (latex) mathematical formula. Then, it becomes evident that this kind of output avoids a very important user action: to copy the output, modify it a little bit, and use it again as a new input. The solution I found for that was to have "InputForm" form as the output by default and, by adding a `//MatrixForm` at the end of an expression, get a MathML output.
> Apart from expressions, there are several other possible outputs. For example, graphics, sounds (from `Play`), 3D graphics, warnings, Print's, HTML, Javascript, etc. which can be produced by the WMA kernel and be handled by the frontend. Jupyter has a rich interface to handle all these kind of objects, and usually you can send several possible outputs and let to the frontend to choose the best choice, according to the available interface (text / web / qt interfaces, for example). The only problem is that you have to build all these different outputs.

When I read this I am both humbled and in awe, but saddened, angered and depressed at the same time.

There are clearly a number of subtleties that have been encountered and worked around. There is a lot of depth to the code. WMA and Mathics have a lot of stuff in there and there is a certain logic to it and the primitives there can be used as the building blocks to make wonderful thing.

That is the part I am in awe and humbled by.

The other part is that, how would one know this short of running into a problem, and investigating.

I don't know or represent that I know what the best solution is. I am pretty sure though that the too-tight integration of SVG, MathML, and the Django front-end has some severe limitations and off gets int the way being able to expand its current capabilities, by offloading the work to other tools and packages.

Instead of taking two building blocks which probably aren't quite right, "InputForm" and "MatrixForm", we should come up with what proper building blocks are.

Right now I am thinking of maybe `SVGForm[ ]` for the SVG _without_ the inline `MathML` and `RsTForm[ ]`  as additional way to mark up help. (Initially conversion at run time of the current markup to RST is straight forward.)

Or is it `ToSVG[ ]`  and `ToRsT[ ]` ? Or Both? Here is where I really don't understand enough about Wofram products to understand what the most natural way to go is.

But I am pretty sure it would not be "InputForm" and "MatrixForm" however clever as a workaround that is.

>
> > Also, I don't get why we treat XML as synonymous for MathML. They are not the same. Or that graphics for TeX is asymptote.
>
> The think is WMA has MathMLForm but not XMLForm to convert expressions. Indeed, the idea was to use XML to format different kind of outputs which can be sent to the Django interface.

While that idea might be workable, its current implementation in Mathics core isn't workable.

When I come across code like this from [_GraphicsElement#boxes_to_xml](https://github.com/mathics/Mathics/blob/master/mathics/builtin/graphics.py#L3198-L3204) which is mixing converting to SVG inside something that purports to handle XML, this feels just wrong. However awesome SVG is, SVG is not the same thing as XML. Maybe I have some other kind of format I'd prefer to render to.

And what I don't get is that Mathics already has its own internal representation. So why do we need XML at all? Why not just stick with that Mathics representation given to the front end and then have the front end to the conversion step. As mentioned in [Mathics Core Interpreter Overview](https://mathics-development-guide.readthedocs.io/en/latest/developing-code/code-overview.html)

>  When you enter a string to Mathics there is a 3-step process:
> * the string is parsed to FullForm
> * The full form S-expression is interpreted
> * The result, also an S-expression, is formatted to the kind of output desired.
>
> This is mentioned from the user perspective in the Figure [“Steps in the operation of Wolfram Language”](https://reference.wolfram.com/language/tutorial/TextualInputAndOutput.html#8971)

The difference is that the _front-end_ does the formatting not Mathics Core.

> Then, you can send a SVG picture, a MathML formula, or any other object, and also you can nest formulas inside graphics, inside formulas.

This we _definitely_ want to preserve. The part that I don't understand is how in the old regime this worked without undoing formatting decisions which is much harder to do, and more unreliable than if the formatting hadn't been done in the first place.

> This is why when you started touching the format for graphics, text output inside MathML stopped working. What we need here is a more clear, robust, and consistent way to get the same behavior.

I'd really like to understand this part better. I'd be grateful if you'd describe how this worked.


> > In particular, we want OutpuForm and FullForm _not_ to produce MathML.
>
> This has two layers: on the one hand, it would be nice to have the possibility of showing beautiful formulas, but also in a way easy to reuse outputs as new inputs.

Here is what I see now with this PR:

```
Graphics[{Rectangle[{1, 1}]}, Axes -> True, PlotRange -> {{-2, 1.5}, {-1, 1.5}}] // OutputForm
```

produces:

```
System`GraphicsBox[System`List[System`RectangleBox[System`List[1, 1]]], System`Rule[System`$OptionSyntax, System`Ignore], System`Rule[System`AspectRatio, System`Automatic], System`Rule[System`Axes, System`True], System`Rule[System`AxesStyle, System`List[]], System`Rule[System`Background, System`Automatic], System`Rule[System`ImageSize, System`Automatic], System`Rule[System`LabelStyle, System`List[]], System`Rule[System`PlotRange, System`List[System`List[-2.0, 1.5], System`List[-1.0, 1.5]]], System`Rule[System`PlotRangePadding, System`Automatic], System`Rule[System`TicksStyle, System`List[]]]
```

If I change "GraphicsBox" to "Graphics" I will get the same output as if I had not entered `// OutputForm` at the end.

I believe there are other cases and situations where this doesn't work as well. And indeed this isn't optimal either. _But it is a start an it is workable whereas before this PR we didn't to my knowledge have something that was workable._

> WMA notebook allows this with machinery that we do not have available by now: It has a native interpreter that builds the formulas, in a way that if you copy a part of it, you get something evaluable.

Here is a suggestion for how to do along the lines started in this PR. Actually, the change was made a little earlier where the returned object has both some sort of formatted object, but it also has the Mathics Expression in `last_result` which doesn't need any scanning or parsing -- it already is in a form which can be evaluated.

> Something less complicated that I thought for IWolfram was to send a MathML expression as an output to the frontend, but with a "onclick" event that sends the "InputForm" expression to the clipboard, or to a new cell.

I am not in a position to judge or say what's better. So let me ask you: how does this compare with the approach mentioned above?

> Graphics have another layer of complexity: on the one hand, they should be able to support formulas inside, for axes labels, captions, etc, ( which is something that asymptote or mpl could handle now, but not 10 years ago).

Yeah a big challenge has been modernizing the code. We have a long way to go still.

> Also, a visual representation of these graphics be built just if the interface support graphics, and in a format that can be used by the specific front-end.
> For all of this, I guess that the best would be to provide support to generate different kind of graphics formats (svg, png) from the mathics kernel, but to control the "makeboxes" for Graphics and Graphics3D on the different frontends.

Note that "kernel" doesn't necessarily mean Mathics Core. It can be something that the Mathics Core calls out to. https://github.com/Mathics3/pymathics-matplotlib has my experiments along this line. By far this needs working out. (Think the state that pymathics was in 6 months ago as opopsed to now.) And it shouldn't use the pymathics mechanism but maybe something more appropriate as a formatting backend.

> Regarding the PR, I think it is OK at this stage at least.

Ok. I will merge as soon as CI okays this.

mmatera:
> When I read this I am both humbled and in awe, but saddened, angered and depressed at the same time.
> 

OK, this is harder to answer, but keep calm, I am sure we can arrive to a much better solution :). 

> There are clearly a number of subtleties that have been encountered and worked around. There is a lot of depth to the code. WMA and Mathics have a lot of stuff in there and there is a certain logic to it and the primitives there can be used as the building blocks to make wonderful thing.
> 
> That is the part I am in awe and humbled by.
> 
> The other part is that, how would one know this short of running into a problem, and investigating?
> 
> I don't know or represent that I know what the best solution is. I am pretty sure though that the too-tight integration of SVG, MathML, and the Django front-end has some severe limitations and gets int the way being able to expand its current capabilities, by offloading the work to other tools and packages.
> 
> Instead of taking two building blocks which probably aren't quite right, "InputForm" and "MatrixForm", we should come up with what proper building blocks are.
> 

The `InputForm`, `OutputForm`, etc were designed for the notebook interface of WMA. Our problem then is to develop our "Forms" keeping the compatibility as high as possible.


> Right now I am thinking of maybe `SVGForm[ ]` for the SVG _without_ the inline `MathML` and `RsTForm[ ]` as additional way to mark up help. (Initially conversion at run time of the current markup to RST is straight forward.)
> 
> Or is it `ToSVG[ ]` and `ToRsT[ ]` ? Or Both? Here is where I really don't understand enough about Wofram products to understand what the most natural way to go is.
> 
Any of them would be OK, if they would be able to provide a consistent output in the corresponding front-end


> But I am pretty sure it would not be "InputForm" and "MatrixForm" however clever as a workaround that is.
> 
I used "MatrixForm" in IWolfram because it existed, and I wanted to keep TeXForm for plain ASCII tex strings, and because it worked out for my usual workflow. Also, when you parse an array by the output of TeXForm, you get exactly a "MatrixForm", so it was not too far from the standard behaviour. On the other hand, I do not use "InputForm" actually. The standard output just consists of convert the expression to String, and send them to jupyter, with a tag that announces that is a plain output.

> > > Also, I don't get why we treat XML as synonymous for MathML. They are not the same. Or that graphics for TeX is asymptote.

Again, is a matter of what WMA keywords use to talk about the objects we need for our frontend.

> > 
> > 
> > The think is WMA has MathMLForm but not XMLForm to convert expressions. Indeed, the idea was to use XML to format different kind of outputs which can be sent to the Django interface.
> 
> While that idea might be workable, its current implementation in Mathics core isn't workable.
> 

OK

> When I come across code like this from [_GraphicsElement#boxes_to_xml](https://github.com/mathics/Mathics/blob/master/mathics/builtin/graphics.py#L3198-L3204) which is mixing converting to SVG inside something that purports to handle XML, this feels just wrong. However awesome SVG is, SVG is not the same thing as XML. Maybe I have some other kind of format I'd prefer to render to.
> 
> And what I don't get is that Mathics already has its own internal representation. So why do we need XML at all? Why not just stick with that Mathics representation given to the front end and then have the front end do the conversion step. As mentioned in [Mathics Core Interpreter Overview](https://mathics-development-guide.readthedocs.io/en/latest/developing-code/code-overview.html)?
> 

The problem is that the browsers do not understand the WL representation to provide a consistent output

> > When you enter a string to Mathics there is a 3-step process:
> > 
> > * the string is parsed to FullForm
> > * The full form S-expression is interpreted
> > * The result, also an S-expression, is formatted to the kind of output desired.
> > 
> > This is mentioned from the user perspective in the Figure [“Steps in the operation of Wolfram Language”](https://reference.wolfram.com/language/tutorial/TextualInputAndOutput.html#8971)
> 
> The difference is that the _front-end_ does the formatting not Mathics Core.
> 
> > Then, you can send a SVG picture, a MathML formula, or any other object, and also you can nest formulas inside graphics, inside formulas.

Exactly

> 
> This we _definitely_ want to preserve. The part that I don't understand is how in the old regime this worked without undoing formatting decisions which is much harder to do, and more unreliable than if the formatting hadn't been done in the first place.
> 

I can answer how this works on IWolfram. The first thing is I prevent to pass to the Wolfram/Mathics formatted using $PreOut.
Then, the expression (without format) is converted to a String by default, in a TeX expression if the expression is enclosed by MathMLForm, in a .wav object if the object is a Sound, in PNG/SVG if the expression is a Graphics object, and in
PNG/javascript string if the expression is a Graphics3D. Then, the output is codified in B64, and proceeded by a tag indicating what kind of output is sent. Certain parameters of the output (using SVG or PNG, active vs pasive 3D graphics) are controlled by  control variables in the Wolfram Kernel.
The IMathics kernel just receive then ASCII strings, which are decoded  and process according to the initial tag.



> > This is why when you started touching the format for graphics, text output inside MathML stopped working. What we need here is a more clear, robust, and consistent way to get the same behavior.
>

In mathics/Django, the thing is different.  Django needs to receive HTML objects. Then, as mathics has this MathMLForm, the output is produced and sent directly. This is the part we need to improve, and I guess the way I used in Iwolfram could work. Again, the main problem here comes from nested expressions, where graphics are part of other expressions, and needs to have expressions inside. MathMLForm did it, because at the end what produced is a nested HTML object. The same happens with TeXForm, since graphics are "rendered" as Asymptote strings, which can contain latex expressions inside.
> 
> I'd really like to understand this part better. I you feel there are advantages to the other approach, I'd be grateful if you'd describe how that worked.
> 
> > > In particular, we want OutpuForm and FullForm _not_ to produce MathML.

I guess the best would be to be able to redefine OutputForm from inside the frontends, which is maybe better than using $PreOut

mmatera:
> > 
> > 
> > This has two layers: on the one hand, it would be nice to have the possibility of showing beautiful formulas, but also in a way easy to reuse outputs as new inputs.
> 
> Here is what I see now with this PR:
> 
> ```
> Graphics[{Rectangle[{1, 1}]}, Axes -> True, PlotRange -> {{-2, 1.5}, {-1, 1.5}}] // OutputForm
> ```
> 
> produces:
> 
> ```
> System`GraphicsBox[
> System`List[System`RectangleBox[System`List[1, 1]]],
> System`Rule[System`$OptionSyntax, System`Ignore],
> System`Rule[System`AspectRatio, System`Automatic],
> System`Rule[System`Axes, System`True],
> System`Rule[System`AxesStyle, System`List[]],
> System`Rule[System`Background, System`Automatic],
> System`Rule[System`ImageSize, System`Automatic],
> System`Rule[System`LabelStyle, System`List[]],
> System`Rule[System`PlotRange,
> System`List[System`List[-2.0, 1.5],
> System`List[-1.0, 1.5]]],
> System`Rule[System`PlotRangePadding,
> System`Automatic],
> System`Rule[System`TicksStyle, System`List[]]]
> ```
> 
> If I change "GraphicsBox" to "Graphics" I will get the same output as if I had not entered `// OutputForm` at the end.
> 
This is because `Graphics` is the "abstract" expression, and `GraphicsBox` should be an intermediate step before converting the expression into a picture.

> There are other cases and situations where this doesn't work as well. And indeed this isn't optimal either. _But it is a start an it is workable whereas before this PR we I don't think what was had was as workable._
> 
> > WMA notebook allows this with machinery that we do not have available by now: It has a native interpreter that builds the formulas, in a way that if you copy a part of it, you get something evaluable.
> 
> Here is a suggestion for how to do along the lines started in this PR. Actually, the change was made a little earlier where the returned object has both some sort of formatted object, but it also has the Mathics Expression in `last_result` which doesn't need any scanning or parsing -- it already is in a form which can be evaluated.

Precisely, I guess that is the step that should be returned by default.

> 
> > Something less complicated that I thought for IWolfram was to send a MathML expression as an output to the frontend, but with a "onclick" event that sends the "InputForm" expression to the clipboard, or to a new cell.
> 
> I am not in a position to judge or say what's better. So let me ask you: how does this compare with the approach mentioned above?

Not sure, I guess it is the same, but I need to understand better how mathics-django is working now.

> 
> > Graphics have another layer of complexity: on the one hand, they should be able to support formulas inside, for axes labels, captions, etc, ( which is something that asymptote or mpl could handle now, but not 10 years ago).
> 
> Yeah a big challenge has been modernizing the code. We have a long way to go still.
> 
> > Also, a visual representation of these graphics be built just if the interface support graphics, and in a format that can be used by the specific front-end.
> > For all of this, I guess that the best would be to provide support to generate different kind of graphics formats (svg, png) from the mathics kernel, but to control the "makeboxes" for Graphics and Graphics3D on the different frontends.
> 
> Note that "kernel" doesn't necessarily mean Mathics Core. It can be something that the Mathics Core calls out to. https://github.com/Mathics3/pymathics-matplotlib has my experiments along this line. By far this needs working out. (Think the state that pymathics was in 6 months ago as opposed to now.) And it shouldn't use the pymathics mechanism but maybe something more appropriate as a formatting backend.
>
In IWolfram, I just use WL to do most of the formatting work in outputs, so in principle, it should not be necessary to use pymathics. However, for mpl, there is an advantage of doing in this way:  mlp handles by itself the interaction with jupyter, and maybe also with django, and handle latex expressions, so using it we can build 2d and 3d graphics with formulas inside, use it to dispatch it to the browser, and also to handle certain interactivity. It is just a matter of translating Graphics expressions into mpl objects.





> 
> > Regarding the PR, I think it is OK at this stage at least.
> 
> Ok. I will merge as soon as CI okays this.

OK

rocky:
@mmatera I've been thinking about this on and off. Basically, you have the most experience here in handling this kind of thing so I think you should take the lead here.

Althoug we want to keep WL compatibility, if IWolfram has workarounds, because it can't change WL, on the Mathics side we should try do design things _without_ the workaround.  Things like using MatrixForm and TeXForm feel like workaround.

mmatera:
Actually, in the command line interface I think mathml is not invoked. What we want to have is the possibility of render mathml and latex from a wl expression. One of my motivations for introducing pymathics was to allow to different front ends to use different low-level make boxes functions.

mmatera:
On the other hand, OutputForm, TraditionalForm and MatrixForm are formats definited in function of an specific graphic front-end. We can redefine them as we prefer, and also we can handle them on the front-end side.

mmatera:
Regarding what is the best default for the render side, and what is the meaning of things like MatrixForm, I am not sure. In Iwolfram the current form is not what was my first idea, and I works for my work flow, but I guess it could be improved a lot.

rocky:
> MatrixForm, I am not sure.

Let me break this down then as I see it. [WL MatrixForm](https://reference.wolfram.com/language/ref/MatrixForm.html) says: 

> prints with the elements of list arranged in a regular array. 
> ...
> In StandardForm the array is shown enclosed in parentheses.

There is nothing in there that I can find that suggests that when you parse an array by the output of TeXForm, you get exactly a MatrixForm or that MatrixForm should not be allowed to print just plain text if it feels like it. 

Indeed MatrixForm is more about the semantics of matrices, not making use of rendering properties of matrices might have, although it is true that when a program needs to render it in some way it will have certain common characteristics. 

> We can redefine them as we prefer, and also we can handle them on the front-end side.

The proposal is to always have rendering initiated by the front end. The routines might come from Mathics core, but the routines might come from elsewhere outside of the Mathics core module.

Although you can redefine the Form functions, it should be done in a way that makes sense. Just because there is a mechanism to define FireEngine to mean the color Red, it probably isn't a good idea to do so. Instead you probabably would want to say that FireEngines generally have a color and the color is generally Red. And you might define a color FireEngineRed. But FireEngines and Colors are still different things. Likewise renderers and Forms are a little different.

rocky:
> but I guess it could be improved a lot.

Ok. What are the failings, and how would you fix those failing in a Mathics environment?

mmatera:
This part is not an issue of mathics, but a limitation of my front end. The ideal think would be to imitate the wma notebook behavior: an formatted output that you can copy and paste in another cell. The following best would be to have a formatted output that you can double click and generate a new cell with the input form of the previous output. Then, we have the present behavior. With it, maybe we can choose better the symbol but do not hgasve in clear which one would be. MatrixForm is just the one that makes the same than the original if you apply it to a matrix. TeXForm and MathMLForm should show the code to paste in the corresponding interpreter. Maybe OutputForm would be the most suitable symbol.... I do not know. Thoughts?

rocky:
I have a hard time understanding or thinking about this in the abstract. So after we get going on the Jupyter interface, I am sure it will be clearer how to handle this. 

In general though with all the flakiness all over the place, I'd rather get more things fixed and more solid before embarking another level of complexity.

PR #9, opened 2021-01-19
Mathics scanner import

rocky:


GarkGarcia (approved):
LGTM but let's wait for https://github.com/Mathics3/mathics-scanner/pull/5 to be merged so that we don't need to install PyYAML in CI here

rocky:
@GarkGarcia Looks like the more complete and agressive substitute everything approach has problems. 

For example: Try 

```
ListLinePlot[{3, 1, 4}]
```

And you'll see on the console: 

```
"GET /[RawDoubleQuote]data:image/svg+xml/[RawSemicolon]base64,CiAgICAg...+CiAgICAgICAg/[RawDoubleQuote]/ HTTP/1.1" 404 55227
```

I think part of the bug is that we do unconditional substitution. But originally that was okay because the _only_ symbols we encoded were the ones that were known to cause problems, not _all_ of them like RawDoubleQuote.

GarkGarcia:
> @GarkGarcia Looks like the more complete and agressive substitute everything approach has problems.
> 
> For example: Try
> 
> ```
> ListLinePlot[{3, 1, 4}]
> ```
> 
> And you'll see on the console:
> 
> ```
> "GET /[RawDoubleQuote]data:image/svg+xml/[RawSemicolon]base64,CiAgICAg...+CiAgICAgICAg/[RawDoubleQuote]/ HTTP/1.1" 404 55227
> ```
> 
> I think part of the bug is that we do unconditional substitution in mathics-django. But originally that was okay because the _only_ symbols we encoded were the ones that were known to cause problems, not _all_ of them like RawDoubleQuote.

I forgot to remove the `\[Raw...]` characters from the maps (I'll fix this right away). I'm not sure about what's going on with the `GET` thing though.

GarkGarcia:
>  think part of the bug is that we do unconditional substitution in mathics-django. But originally that was okay because the _only_ symbols we encoded were the ones that were known to cause problems, not _all_ of them like RawDoubleQuote.

Yes, we should definitively not substitute graphics.

GarkGarcia:
> > @GarkGarcia Looks like the more complete and agressive substitute everything approach has problems.
> > For example: Try
> > ```
> > ListLinePlot[{3, 1, 4}]
> > ```
> > 
> > 
> > And you'll see on the console:
> > ```
> > "GET /[RawDoubleQuote]data:image/svg+xml/[RawSemicolon]base64,CiAgICAg...+CiAgICAgICAg/[RawDoubleQuote]/ HTTP/1.1" 404 55227
> > ```
> > 
> > 
> > I think part of the bug is that we do unconditional substitution in mathics-django. But originally that was okay because the _only_ symbols we encoded were the ones that were known to cause problems, not _all_ of them like RawDoubleQuote.
> 
> I forgot to remove the `\[Raw...]` characters from the maps (I'll fix this right away). I'm not sure about what's going on with the `GET` thing though.

Ok, I've fixed the `\[RawDoubleQuote]` in https://github.com/Mathics3/mathics-scanner/commit/29f87d258d6b362d30d44af28ee4f7322eaba05a but it looks like the GitHub action for auto-compiling the JSON file isn't really working.

GarkGarcia:
> Ok, I've fixed the `\[RawDoubleQuote]` in [Mathics3/mathics-scanner@29f87d2](https://github.com/Mathics3/mathics-scanner/commit/29f87d258d6b362d30d44af28ee4f7322eaba05a) but it looks like the GitHub action for auto-compiling the JSON file isn't really working.

This issue was a bug in `mathics-scanner`.

GarkGarcia:
@rocky I've fixed the `\[RawDoubleQuote]` issue now.

GarkGarcia:
It looks like https://github.com/Mathics3/mathics-scanner/commit/29f87d258d6b362d30d44af28ee4f7322eaba05a was enough to fix the issue, but we should definitively check for graphics before calling `replace_wl_with_plain_text`.

mmatera:
Take into account that inside graphics, you have labels that could have symbols inside...

mmatera (approved):
LGTM

GarkGarcia:
> Take into account that inside graphics, you have labels that could have symbols inside...

Humm, you have a point. Well, this is something we need to consider when redesigning graphics: if they aren't just URIs for SVG images it may be easier to manipulate to strings inside them. 

Alternatively, we could implement a routine that uses `replace_wl_with_plain_text` to recursivelly replace the text inside an expression _only when the replacement is needed_ (in symbols and strings I guess). This routine cannot be implemented in `mathics-scanner`, since we need to go over the leaves of the ADT representation of the expression.

GarkGarcia:
> Alternatively, we could implement a routine that uses `replace_wl_with_plain_text` to recursivelly replace the text inside an expression _only when the replacement is needed_ (in symbols and strings I guess). This routine cannot be implemented in `mathics-scanner`, since we need to go over the leaves of the ADT representation of the expression.

This option is probably more robust overall. Even if this problem can be delegated to graphics redesign, replacing _everything_ as we're doing will likely cause other issues.

rocky:
> Humm, you have a point. Well, this is something we need to consider when redesigning graphics: if they aren't just URIs for SVG images it may be easier to manipulate the strings inside them.

The bug here is not in graphics or redisigning graphics. It is in how the formatting routines work. They need to know whether the preferred output is to use unicode or replace Unicode symbols with character names or equivalent characters like `->` for arrow. 

> Alternatively, we could implement a routine that uses replace_wl_with_plain_text to recursivelly replace the text inside an expression only when the replacement is needed (in symbols and strings I guess).

The existing format routines in fact do traverse expressions 

> This routine cannot be implemented in mathics-scanner, since we need to go over the leaves of the ADT representation of the expression.

Correct. The formatter though still needs the maps to understand when a translation is needed and what it is.

PR #25, opened 2021-02-21
setting using asymptote for graphics by defailt

mmatera:


setup.py
```diff
@@ -49,7 +49,7 @@ def read(*rnames):
 is_PyPy = platform.python_implementation() == "PyPy"
 
 INSTALL_REQUIRES = []
-DEPENDENCY_LINKS = []
+DEPENDENCY_LINKS = ["http://github.com/Mathics3/pymathics-asy#egg=pymathics-asy"]
```
rocky, line 52: Please don't do this. We want to reduce complexity and the number of dependencies and this just adds this back in.

The front ends should allow for reading profiles and customization. 

And the mathics-ominbus project was created to pull in all of the different pieces.

mmatera:
@rocky , If you can, I would appreciate some help in setting the dependency for the module in setup.py

rocky:
> @rocky , If you can, I would appreciate some help in setting the dependency for the module in setup.py

Ok I will look at over the weekend.

setup.py
mmatera: OK, then I would discard this PR. However, I think this would be a suitable (temporary) patch to fix the issue with graphics, this is why I put it here. 
The other way to solve the graphics issue would be to update and debug all the javascript machinery.

rocky: We should probably think about and discuss a strategy, short-term and long term. 

I confess that in a way I did I think the same kind of thing with pymathics-graphics that you are doing here. So I understand and am sort of sympathetic to short-term workarounds. 

Personally, I am happier with short-term workarounds only after we have a clear idea of what the longer-term strategy is. 

A long while ago, it was my impression that we were going to eventually ditch Django and go with some sort of Jupyter-based solution. At least in my mind that hasn't been feasible because the foundations of the Mathics core and how it interacts with "front-ends" just doesn't feel fully worked out, although we are in my opinion making good progresss toward that end,.

mmatera: OK, I agree with that, and I thought that the strategy was more or less this one:
* Fix graphics in Django with a plugin (pymathics-matplotlib or this one) that works in a very short term, in order to have a functional candidate for release soon. I would also prefer to use Matplotlib instead Asymptote for rendering graphics, but it happens that the development of the asy interface is much more advanced than the mpl one. This is why I have used it in this PR.

*  Use this to improve the way in which graphics packages are loaded and works. Regarding this, working on pymathics-asy I think I have discovered the root of the problem with graphics: it is rooted in the javascript/MathJax postprocessing. To really fix it we should look at the javascript routines that perform the XHTML replaces (mathics.js). We will need to do that if we want to have svg graphics. By now, I solve this by sending png graphics.

*  In the long term,  to use what we learned working on that plugin to improve the jupyter's interface, which in the long term replace Django frontend as the main frontend of the project. Yesterday I have been doing some progress in that direction. I am updating this binder: [https://mybinder.org/v2/gh/mmatera/nbcoursebase/master](https://mybinder.org/v2/gh/mmatera/nbcoursebase/master) with the improvements to make online checks.



Yesterday, I was doing some progress in both points: 
* I think I have discovered the root of the issues with graphics in django: MathJax/mathics.js overwrites the image or the text in the image. To really fix it and allow svg graphics, we need to fix the javascript auxiliary routines. The draconian solution I propose is then to use png images.
*  
*   
*

rocky: Ok. Thanks for the information and plan! 

Let me think about abount this, look at the code, try things, and try to understand the what's been done and the existing code.

PR #26, opened 2021-03-07
Delete worksheets

danielpyon:
This adds deletion of worksheets. The user simply needs to click "delete" in the popup:

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,5 @@ mathics_django/web/media/doc/classes.pdf
 mathics_django/web/media/doc/classes.png
 mathics_django/web/media/pdf/
 tmp
+env/
+venv/
--- a/mathics_django/web/media/js/inout.js
+++ b/mathics_django/web/media/js/inout.js
@@ -43,7 +43,7 @@ function showWorksheets(isOpen) {
 				));
 			});
 
-			// If the popup is not open, open it
+			// If the popup is not open, open it.
 			if (!isOpen)
 				showPopup($('open'));
 		}
--- a/mathics_django/web/media/js/inout.js
+++ b/mathics_django/web/media/js/inout.js
@@ -30,12 +30,12 @@ function showWorksheets(isOpen) {
 			tbody.deleteChildNodes();
 			response.worksheets.each(function(worksheet) {
 				tbody.appendChild($E('tr',
-					$E('th',
+					$E('td',
 						$E('a', {'href': 'javascript:openWorksheet("' + worksheet.name + '")'},
 							$T(worksheet.name)
 						)
 					),
-					$E('th',
+					$E('td',
 						$E('a', {'href': 'javascript:deleteWorksheet("' + worksheet.name + '")'},
 							$T("Delete")
 						)
--- a/mathics_django/web/media/js/inout.js
+++ b/mathics_django/web/media/js/inout.js
@@ -21,24 +21,51 @@ function openWorksheet(name) {
 	})
 }
 
-function showOpen() {
-	requireLogin("You must login to open online worksheets.", function() {
-		new Ajax.Request('/ajax/getworksheets/', {
-			method: 'get',
-			onSuccess: function(transport) {
-				var response = transport.responseText.evalJSON();
-				var tbody = $('openFilelist');
-				tbody.deleteChildNodes();
-				response.worksheets.each(function(worksheet) {
-					tbody.appendChild($E('tr', $E('td',
+function showWorksheets(isOpen) {
+	new Ajax.Request('/ajax/getworksheets/', {
+		method: 'get',
+		onSuccess: function(transport) {
+			var response = transport.responseText.evalJSON();
+			var tbody = $('openFilelist');
+			tbody.deleteChildNodes();
+			response.worksheets.each(function(worksheet) {
+				tbody.appendChild($E('tr',
+					$E('th',
 						$E('a', {'href': 'javascript:openWorksheet("' + worksheet.name + '")'},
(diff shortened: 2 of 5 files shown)


rocky:
@danielpyon overall I like this. Thanks for handling! 

The CI failure we are getting seems to be a failing in a change made to the Mathics side recently. I have adjusted things there. 

However I am not currently seeing a "delete" button although I do see the worksheets I have: 

mathics-django


One other thing you might consider is describing this in the help system: 

django-help

But that woul be done on the Mathics repo side (for now until we get convert that to live here.) That would be in https://github.com/mathics/Mathics/blob/master/mathics/doc/documentation/1-Manual.mdoc#L987

rocky (changes requested):
@danielpyon overall I like this. Thanks for handling!

However I am not currently seeing a "delete" button although I do see the worksheets I have:



The CI failure we are getting seems to be a failing in a change made to the Mathics side recently. I have adjusted things there. It looks like though you need to update this PR to have the adjusted code to kick in. But if you fix the problem I am seeing with the delete button not showing up, then I think this will fix the CI problem: 

mathics-django

rocky:
@danielpyon Hmm.... more information. That screenshot comes from google chrome. If I use firefox or opera, the delete button does appear.

danielpyon:
That's strange. For my machine, it works just fine. I will have to some more investigation.

On a related note, the documentation doesn't seem to contain some of the functions like ParametricPlot. Additionally, the Import function does not work on my machine. This is why I think it may be a problem on my end, although I'm not sure what, since I just cloned and built the repository.

danielpyon:
> @danielpyon Hmm.... more information. That screenshot comes from google chrome. If I use firefox or opera, the delete button does appear.

Ah, try refreshing your cache with Ctrl + F5?

rocky:
> > @danielpyon Hmm.... more information. That screenshot comes from google chrome. If I use firefox or opera, the delete button does appear.
> 
> Ah, try refreshing your cache with Ctrl + F5 (in Chrome)?

Thanks! - that worked!

rocky:
> On a related note, the documentation doesn't seem to contain some of the functions like ParametricPlot. 

It might be because you haven't built the data for the doc on the Mathics side (which takes a bit of time). 

In the Mathics repo, right `make doc-data`.  and you should see a large file `<root-mathics-github>/Mathics/mathics/doc/xml/data`  .
> Additionally, the Import function does not work on my machine. This is why I think it may be a problem on my end, although I'm not sure what, since I simply cloned and built the repository.

Can you get help on the `Import` builtin? : 

```
In[1]:= ?Import

          'Import["file"]'
            imports data from a file.
          'Import["file", elements]'
            imports the specified elements from a file.
          'Import["http://url", ...]' and 'Import["ftp://url", ...]'
            imports from a URL.
```

rocky (approved):

danielpyon:
> It might be because you haven't built the data for the doc on the Mathics side (which takes a bit of time).

That makes sense. How would I do this in a virtualenv?

image

Help works, but for instance, the images aren't actually displayed.

rocky:
> > It might be because you haven't built the data for the doc on the Mathics side (which takes a bit of time).
> 
> That makes sense. How would I do this in a virtualenv?
> 
> image
> 
> Help works, but for instance, the images aren't actually displayed.
> 
> It's probably because I haven't gotten the latest Mathics version or something like that.

When you build from git _source_ you have to build the documentation. There is a GNU Makefile in the Mathics to simplify this and that is what `make doc-data` does. The underlying command is: 

```
python mathics/test.py -ot -k 
````

so you could run that from a command line of the Mathics source as well.

virtualenv is just about where python and its packages are installed. It is totally separate from GNU make and only incidentally related to running that command above. 

When you download a package, the package should have the extracted doc inside the package. For example: 

```
$ tar -tf  /mathics/Mathics/dist/uploaded/Mathics3-2.0.0.tar.gz | grep xml/data
Mathics3-2.0.0/mathics/doc/xml/data
```

**Comment by danielpyon** [comment]
Whoops, yeah I was confused there. I forgot to run ```make``` for Mathics.

**Comment by rocky** [comment]
@danielpyon FYI We are very weak on people contributing to the Django front-end which is why it isn't that featureful.

Here are someome weaknesses, in case you can help out...

* updating mathjax to a version that is newer than 6 years old. 
* figuring out why https://github.com/mathics/Mathics/pull/1145 is broken on google chrome and opera, but not on firefox.

**Comment by danielpyon** [comment]
I will definitely check those out, though I was hoping to add autocomplete soon. Is it okay if I work on that first and then look at those weak points?

If autocomplete isn't high priority then I suppose I could just help with those issues directly.

**Comment by rocky** [comment]
Of course since you are doing the work you can work on whatever interests you.

Really good autocomplete is tricky. However there is some in mathicscript. See https://github.com/Mathics3/mathicsscript/blob/master/mathicsscript/termshell.py#L306-L362

**Comment by danielpyon** [comment]
Is there a way to get just the names of all functions from the documentation?

I made a basic prototype [here](https://codepen.io/danielpyon/details/NWbOORB), which requires a list of all names to search for. I'm planning on having two data structures: one unchanging trie that contains all the builtin functions/constants and loads once on startup, and another one that dynamically changes every time the user defines something. That way, the autocomplete suggestions will include the locally declared stuff as well as the default stuff.

**Comment by rocky** [comment]
Look at the Definitions class in Mathics/mathics/core/definitions.py and `get_names()`: 

```
    def get_names(self):
        return (
            self.get_builtin_names()
            | self.get_pymathics_names()
            | self.get_user_names()
```

PR #27, opened 2021-03-11
Basic autocomplete

danielpyon:
This PR adds an autocomplete box to the top right corner. The box opens when the user starts typing, and they can scroll through suggestions with ```Alt``` and select one with ```Escape```.

Things left to improve:
* Explain to the user how to use autocomplete
* Update list of words every time something is assigned
* Use down arrow and ```Tab``` instead of ```Alt``` and ```Escape```
* Better styling

I tried to get the box to follow the cursor, but it was 1) sluggish and 2) required janky calculations involving font width, which I didn't end up figuring out.

### DIFF
--- a/mathics_django/web/media/css/styles.css
+++ b/mathics_django/web/media/css/styles.css
@@ -714,3 +714,13 @@ div.filelist table a {
 div.filelist table a:hover {
   color: #33c;
 }
+
+#dropdown {
+  background-color: lightgray;
+  width: 300px;
+  display: none;
+  position: fixed;
+  top: 0px;
+  right: 0px;
+  font-family: monospace;
+}
--- a/mathics_django/web/media/js/autocomplete.js
+++ b/mathics_django/web/media/js/autocomplete.js
@@ -162,6 +162,7 @@ function autocomplete(e) {
 			e.preventDefault();
 			let replacement = candidates[idx];
 			e.target.value = hungryReplace(text, replacement, pos - 1);
+			dropdown.style.display = 'none';
 		}
 	} else {
 		idx = 0;
--- a/mathics_django/web/media/js/autocomplete.js
+++ b/mathics_django/web/media/js/autocomplete.js
@@ -1,0 +1,190 @@
+
+function TrieNode(key) {
+	this.key = key;
+	this.parent = null;
+	this.children = {};
+	this.end = false;
+}
+
+TrieNode.prototype.getWord = function() {
+	var output = [];
+	var node = this;
+	while (node !== null) {
+		output.unshift(node.key);
+		node = node.parent;
+	}
+	return output.join('');
+};
+
+function Trie() {
+	this.root = new TrieNode(null);
+}
+
+// time complexity: O(k), k = word length
+Trie.prototype.insert = function(word) {
+	var node = this.root;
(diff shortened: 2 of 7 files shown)


rocky:
Right now I don't think this is ready. 

It took me quite a while to figure out how to use, even with the instructions above. I hadn't realized that the keys `Alt` or `Escape` only work in the input area not the dropdown area. When I saw the completion dropdown I went to that box and started using the dropdown. In that area neither `Alt` nor `Escape` and most (all?) of the other keys do nothing.  And when there is only one completion, having a dropdown offered that does nothing but select the select item shown isn't helpful. 

Later, the dropdown arrow disappeared _except_ when there is only one completion and I have no idea why. 

For what we have right now this isn't all that different from the completion that is in help already which is also in that same area. Having the two do almost the same thing, in my opinion, adds to the clutter and confusion

Maybe there is a way to hook the two together so that when in the help area some there is a key to copy the documented function back into the input area? 

As you mention user-defined variables aren't handled. Also the context isn't handled. If I type `\[`  in the input area, then named-character words like Alpha or Pi should be offered. That's what the command-line interface  `mathicsscript` does. And if I start to type a quote, then I have started writing a string and completion of function names doesn't help often. Again `mathicsscript` handles this. 

In general what's going on is that to do good completion you need to also take into account the text that comes before. 

It is not so much that `mathicsscript` is clever except that it uses an _existing command-completion framework_ that has already thought about and solved these problems. When you roll your own system you have to rediscover and solve all of these things.

Personally, I'd like to see some of the existing capabilities in Mathics and the Django front end worked on get better and more solid, rather than add more incomplete features, which when released will just cause us to get questions over the use.

In particular, the last PR which allows for workspaces to be deleted, could use some documentation describing how that works. If you want to make that slicker, using a checkbox of items to delete with a submit buttion prompting for confirmation seems more natural than a link with no confirmation before deletion occurs.

In the big picture, we'd like to move away from Django and toward a Jupyter interface. That is what people probably would be more familiar with and that already comes with features for saving worksheets, renaming and deleting them so we don't have to spend time duplicating this effort. 

Of course we'll have to keep Django around for a while, if not forever. So spending a little time on that is warranted if people get immediate benefit from a little bit of work. But spending a lot of time to recreate a user interface that probably exists already in that framework feel to me like it is drawing effort away from the places where our limited resources we have would be better put to use.

mmatera:
@danielpyon , maybe this could be helpful:
[(https://github.com/mmatera/iwolfram/blob/2df3673967e745ff621b57f04bd212c72314cfa0/wolfram_kernel/wolfram_kernel.py#L1232](https://github.com/mmatera/iwolfram/blob/2df3673967e745ff621b57f04bd212c72314cfa0/wolfram_kernel/wolfram_kernel.py#L1232)
Of course this comes from the other project based on jupyter, but the idea is the same...

mmatera:
and [here](https://pypi.org/project/django-completion/) there is a project that implements self-completion in Django. If you really want to do this, maybe this deserves a look

danielpyon:
Yeah that makes sense. I didn't know that the interface would switch to Jupyter. In that case, I agree that this is far from being usable, and probably isn't worth continuing.

PR #38, opened 2021-05-01
About

rocky:
Add rudimentary "about" url `http://localhost:8000/about`, htttp://localhost/about.html`, htttp://localhost/about.htm` 

which prints what version of Django, MathJax, three.js and Mathics-version info that is getting used.

Unless there is interest or objections in using, I'll merge over the weekend.


mmatera (approved):
LGTM

PR #39, opened 2021-05-11
Don't use MathJax for string output

rocky:
This is minimal code. Down the line I'd like to put this under a setting, specific to Django. 

But doing this properly means redoing how settings work in Mathics. The Django-specific setting there in `Settings.m` should be moved here. 

But doing this properly means pulling out the autoload part of of Definitions and putting that into its own function.  Sigh.

### DIFF
--- a/mathics_django/web/format.py
+++ b/mathics_django/web/format.py
@@ -60,8 +60,11 @@ def format_output(obj, expr, format=None):
         # This part is custom to mathics-django:
         if str(expr) == "-Graph-" and hasattr(expr, "G"):
             return format_graph(expr.G)
-        elif str(expr.get_head()) == 'System`CompiledFunction':
+        head = str(expr.get_head())
+        if head == 'System`CompiledFunction':
             result = expr.format(obj, "System`OutputForm")
+        elif head == 'System`String':
+            result = expr.format(obj, "System`InputForm")
         else:
             result = Expression("StandardForm", expr).format(obj, "System`MathMLForm")
     else:


mmatera:
Regarding the output format, you can also use the IOHooks to force that (for example) strings appears in InputForm. BTW, why do you want to get this behavior?

rocky:
I want String text to be visually distinct. Also it makes no sense to have string output be in MathML mode. 

Down the line I probably also want HREFs for matplotlib images (and other images) which also shouldn't be in MathML mode.


> Regarding the output format, you can also use the IOHooks to force that (for example) strings appears in InputForm. BTW, why do you want to get this behavior?

I want String text to be visually distinct. Also it makes no sense to have string output be in MathML mode. 

Down the line I probably also want HREFs for matplotlib images (and other images) which also shouldn't be in MathML mode.

rocky:
> > > Regarding the output format, you can also use the IOHooks to force that (for example) strings appears in InputForm. BTW, why do you want to get this behavior?
> > 
> > 
> > I want String text to be visually distinct. Also it makes no sense to have string output be in MathML mode.
> > Down the line I probably also want HREFs for matplotlib images (and other images) which also shouldn't be in MathML mode.
> 
> OK; but this is something that should be implemented by redefining `MakeBoxes` (or `ToBoxes`? or ToString?) for strings. The problem is that usually you have nested expressions (strings inside list, strings inside Graphics, etc..) the `format` method looks at these definitions, which allows to control that from the front-end interface.

Yes, that is a problem, but I think a _separate_ problem. At  the top-level I don't want MathML all of the time.  Front-ends  have the know what is most appropriate for the object it gets back. String is one case, Graphics3D may be another NetworkX graphics may be another.

PR #40, opened 2021-05-12
Update to MathJax 2.3

rocky:
If we manage to get to 2.5 supposedly there is a big performance boost. 

But this is a stepping stone along the way it get us to circa 2013.


mmatera (approved):
It passes both the tests and the demo, both in Firefox and Chrome, so let's merge it.

PR #41, opened 2021-05-12
Add ability to autoload Mathics ...

rocky:
used to set settings.


rocky:
This will probably remain broken until https://github.com/mathics/Mathics/pull/1369 gets into master and/or some github magic for pulling this version gets added.

PR #43, opened 2021-05-14
Upgrade to MathJax 2.5.0

rocky:
Pulls in way too many font files, but this is an important useful release. The bloat and having a cleaner way to install MathJax will be addressed later.


mmatera:
Would be possible to put all of this in a tar file, and then make that the installer decompress it?

rocky:
> Would be possible to put all of this in a tar file, and then make that the installer decompress it?

Wheel and egg formats _are_ basically compressed archive files of some sort which are uncompressed on install.

When I look at MathJax 2.7.4 I am not seeing `fonts/TeX` around which is the bulk of this. So if we ever get there this aspect will be solved by itself without our doing any work.

mmatera (approved):
In that case, go ahead...

rocky:
> In that case, go ahead...

Thanks - after release in the plan is to move forward in releases. This is just yet another intermediate step.

rocky:
@mmatera I am seeing in chrome, but not firefox a bar after each equation. I haven't been able to figure out where that is coming from though 
chrome-extraneous bar

mmatera:
Does this happen after the MathJax update? It seems to be a kind of a cursor mark. Could you do a right-click over the equation, get the MathML code that Chrome and Firefox are showing, and paste them here?

rocky:
> Does this happen after the MathJax update? It seems to be a kind of a cursor mark. Could you do a right-click over the equation, get the MathML code that Chrome and Firefox are showing, and paste them here?

Yes, it is definitely 2.5.0 related. And in the chrome inspector it is not seen in the equation but in the MathJaxDisplay div that immediately surrounds it. Have to postpone looking at this for now.

mmatera:
I think that could worth to leave this as it is now, and at some point upgrade MathJax to the next version, to see if this still happens.

PR #48, opened 2021-05-25
Django doc

rocky:


mathics_django/doc/django_doc.py
```diff
@@ -0,0 +1,658 @@
+# -*- coding: utf-8 -*-
```
mmatera, line 1: Is there any difference between this and `mathics.doc.doc` ? Wouldn't be possible to just import the mathics (core) version?

mmatera:
I was having a glance at this.  So, we need `mathics_django.doc.doc` because we want versions of the objects in `mathics.doc.doc` with a `get_url` method? Or the idea is to split off the mathics.doc.doc into a latex module and an HTML module?

rocky:
> I was having a glance at this. So, we need `mathics_django.doc.doc` because we want versions of the objects in `mathics.doc.doc` with a `get_url` method? Or the idea is to split off the mathics.doc.doc into a latex module and an HTML module?

Split off into modules outside of Mathics-Django and Mathics core. However doing this all in one step is too difficult.

Recall that just splitting off Django was difficult. And as a result of that the doc.py got duplicated. And then over time they are out of date. And the duplication is not just ugly and makes things more complicated than it needs, but also is a maintenance problem.

mmatera:
@rocky, I do not mean to split doc as a separated project, but inside `mathics.doc`. The idea would be to do not spread the `doc` code over different projects.

rocky:
> @rocky, I do not mean to split doc as a separated project, but inside `mathics.doc`. The idea would be to do not spread the `doc` code over different projects. Once this is achieved, then it should be simpler to move this as a separate project.

If you look at the django-related HTML code it wants to use django routines for escaping things (as it should). 

I would like to see this going to sphinx with autodoc and that will pull in sphinx.

mmatera:
> > @rocky, I do not mean to split doc as a separated project, but inside `mathics.doc`. The idea would be to do not spread the `doc` code over different projects. Once this is achieved, then it should be simpler to move this as a separate project.
> 
> If you look at the django-related HTML code it wants to use django routines for escaping things (as it should).
> 
> I would like to see this going to sphinx with autodoc and that will pull in sphinx.
> 
> The module outside could be called mathics-doc.

Do you mean line 23 in mathics_django/doc/django_doc.py ?
`from mathics_django.doc.utils import escape_html, slugify`

Probably this is not django code neither... But OK, I agree that once the splitting becomes clearer, we want a mathics-doc package.

rocky:
> > > @rocky, I do not mean to split doc as a separated project, but inside `mathics.doc`. The idea would be to do not spread the `doc` code over different projects. Once this is achieved, then it should be simpler to move this as a separate project.
> > 
> > 
> > If you look at the django-related HTML code it wants to use django routines for escaping things (as it should).
> > I would like to see this going to sphinx with autodoc and that will pull in sphinx.
> > The module outside could be called mathics-doc.
> 
> Do you mean line 23 in mathics_django/doc/django_doc.py ?
> `from mathics_django.doc.utils import escape_html, slugify`

No. I mean https://github.com/Mathics3/mathics-django/blob/493c46b94a7d80ceb8a7238994b8e39d5c7d1579/mathics_django/doc/django_doc.py#L15

I would not be surprised if we could reduce or eliminate `escape_html`  making use of things from Django's libraries or one of the libraries it uses. In fact I might have had more of that in there at one time and had to remove it because of trouble in fully trying to modularize the code. 

Perhaps in 2006 the code may have thought of itself as being a "lightweight" and an expert on Documentation, Web clients, Command-line Clients which therefore would dip into  Django, HTML, LaTeX,  SVG, MathML, threejs NLP and so on. 

However in 2021 this approach has is holding us back in scaling things. And we too often have reinvented what is better done outside. 

Also, all of the strong coupling and homegrown code in several thousand-line modules a nightmare has, for me, made the code hard to work with and understand. It has made testing things the integrated pieces hard to test in isolation while making the entire testing process take a long time.


> 
> Probably this is not django code neither... But OK, I agree that once the splitting becomes clearer, we want a mathics-doc package.

PR #50, opened 2021-05-31
First attempt to set tick style colors

rocky:
@TiagoCavalcanteTrindade Here is a simple attempt to allow TicksStyle to get understood in ThreeJS. 

First, we assume (for now) that TicksStyle is only a color. There can be 0..3 of them. If a color doesn't appear, it uses "black". 

What do you think? I am sure there are ways to improve it.

### DIFF
--- a/.github/workflows/osx.yaml
+++ b/.github/workflows/osx.yaml
@@ -24,6 +24,9 @@ jobs:
         brew install llvm@11
         python -m pip install --upgrade pip
         LLVM_CONFIG=/usr/local/Cellar/llvm@11/11.1.0/bin/llvm-config pip install llvmlite
+        # Can remove when next Mathics is released
+        python -m pip install -e git://github.com/mathics/Mathics#egg=Mathics3
+        pip install -e .
     - name: Install Mathics Django
       run: |
         make develop
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -22,6 +22,9 @@ jobs:
       run: |
         sudo apt-get update -qq && sudo apt-get install -qq liblapack-dev llvm-dev
         python -m pip install --upgrade pip
+        # Can remove when next Mathics is released
+        python -m pip install -e git://github.com/mathics/Mathics#egg=Mathics3
+        pip install -e .
     - name: Install Mathics Django
       run: |
         make develop
--- a/.github/workflows/windows.yaml
+++ b/.github/workflows/windows.yaml
@@ -25,6 +25,8 @@ jobs:
         python -m pip install wheel
         choco install llvm
         set LLVM_DIR="C:\Program Files\LLVM"
+        # Can remove when next Mathics is released
+        python -m pip install -e git://github.com/mathics/Mathics#egg=Mathics3
         pip install -e .
     - name: Install Mathics Django
       run: |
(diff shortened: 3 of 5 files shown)


TiagoCavalcanteTrindade:
@rocky I think it's a good idea.

And about updating tree.js, I'm stuck in drawing the axes, but all drawing functions except cube are working.

rocky:
> @rocky I think it's a good idea.
> 
> And about updating three.js, I'm stuck in drawing the axes, but all drawing functions except cube are working.

I really don't know that much about javascript,  three.js and its 3D graphics model. That's why I opted for going forward more ore less incrementally in three.js releases. The hope was that the code would grow and adapt in the same way that three.js has.

So a suggestion is to try a release that is not too far from r52 and get that working. Of course, parts of the rewritten code might be useful there. And as you get closer to the last release presumably all of it (except the parts that have problems) would.

TiagoCavalcanteTrindade:
@rocky, I'll have some time next week and I'll help in update to versions closer to r52.

I suggested to update to the newest version because of with a big change the code will be rewrote by almost whole, and [if we could do it] the code will be better (not offending, but it isn't too close to good).

rocky:
> @rocky, I'll have some time next week and I'll help in update to versions closer to r52.
> 
> I suggested to update to the newest version because of with a big change the code will be rewrote by almost whole, and [if we could do it] the code will be better (not offending,

Sure, I understand. I also appreciate your putting in the work to try this. And I sincerely hope that we will ultimately make use of this.  but it isn't too close to good).

Many times it happens that I try the same kind of thing myself. If I am lucky it works!

More often though I am not so lucky.  And in those cases, I have to take a step back. Or sometimes letting the work sit for a while, I can figure out what I did wrong or what approach I can take to fix things. 

Again, my heartfelt thanks for tackling this long-standing problem.

rocky:
@TiagoCavalcanteTrindade Are you up for adapting this for the new code associated with r54? 

(I have just revised master so that we the failures which have nothing to do with this code are gone.)

TiagoCavalcanteTrindade:
@rocky I'll do that. Just give me 20 minutes.

TiagoCavalcanteTrindade:
@rocky I'm having the following error after update Mathics (this was necessary to access axes colors).

```py
ImportError: cannot import name 'autoload_files' from 'mathics.core.definitions' (/home/tiago/.local/lib/python3.9/site-packages/mathics/core/definitions.py)
```

rocky:
> @rocky I'm having the following error after update Mathics (this was necessary to access axes colors).
> 
> ```python
> ImportError: cannot import name 'autoload_files' from 'mathics.core.definitions' (/home/tiago/.local/lib/python3.9/site-packages/mathics/core/definitions.py)
> ```

Look in `home/tiago/.local/lib/python3.9/site-packages/mathics/core/definitions.py` to see if this is the right file and indeed there is  indeed `def autoload_files` on line 44. 

A common Python thing is getting the wrong module loaded. If for example you trying to run from source, the above says that's not going on here but instead you are getting a Python 3.9 mathics that had been installed previously.

TiagoCavalcanteTrindade:
@rocky no, this is the line 43: `class Definitions(object):`, ad the 44: `def __init__(self, add_builtin=False, builtin_filename=None):`

> A common Python thing is getting the wrong module loaded. If for example you trying to run from source, the above says that's not going on here but instead you are getting a Python 3.9 mathics that had been installed previously.
What I should do to correct this?

TiagoCavalcanteTrindade:
@rocky I uninstalled and reinstalled some packages and now it's working.

PR #52, opened 2021-06-01
Fix JS pxpx bug ...

rocky:
We were getting "pxpx" in Javascript console if the width or height
already had px in it.

Todo: Could generalize to check for a general suffix by checking
integerness of width and height.


mmatera (approved):
LGTM

PR #47, opened 2021-05-17
Add color to axes numbers in 3D plot

TiagoCavalcanteTrindade:
I added colors to the axes numbers in 3D plot with the same color scheme than Sketchup (x: red, y: green, z: blue) and also replaced `0.` by `.` in the  axes numbers of 3D plot for better visualization.

example of 3D graphic after my changes

PR #56, opened 2021-06-05
Update threejs to r54

TiagoCavalcanteTrindade:
This update three.js from r52 to r54 and fix the drawing polygon bug mentioned in #47.

@rocky I also tested the bug in #47 for other polygons and the new version drew what was expected.


rocky:
@TiagoCavalcanteTrindade  Many thanks!

PR #60, opened 2021-06-06
Implement TicksStyle

TiagoCavalcanteTrindade:
@rocky I used your code to implement `TicksStyle`.

If `TicksStyle` don't be specified or an empty list is passed to it the color will be black.  
If only one color is passed (e.g.: `TicksStyle->Green`) all axes will be of that color.  
If a list of 1, 2 or 3 elements is passed the firsts axes will be of these color.  
If a list of more than 3 elements is passed the color will be black.


rocky:
@TiagoCavalcanteTrindade  Looks great - thanks!

Issue #62, opened 2021-06-07
PointSize is not handled in 3D


rocky:
Try 
```
Table[Graphics3D[{PointSize[r], Point[{0, 0, 0}]}], {r, {0.02, 0.05, 
   0.1, 0.3}}]
```

Recently in https://github.com/mathics/Mathics/pull/1411 field member `pointSize` was added to the [data that is passed in json](https://github.com/mathics/Mathics/blob/b8968a0c3a6ffb1f34484d4c31d6b0068888a450/mathics/formatter/json.py#L73). 

@TiagoCavalcanteTrindade is this somethone you'd be interested in? 

(There are also problems in aymptote for 2D and PlotStyle, but those will be handled separately in the Mathics core project.

TiagoCavalcanteTrindade:
@rocky I think it'll be easy to do.

rocky:
@TiagoCavalcanteTrindade  Wow - that was fast. However the pionts look like black squares (not boxes because they don't rotate) rather than round points. 

Could that be addressed? 

![pointsize](https://user-images.githubusercontent.com/8851/121052364-4e6a4880-c788-11eb-82f9-4cbe367f45f3.png)

TiagoCavalcanteTrindade:
@rocky I will see if I can fix this.

rocky:
Fixed by #63

PR #63, opened 2021-06-07
Implement PointSize for 3d graphics

TiagoCavalcanteTrindade:
This change implements `PointSize` for 3d graphics.

### DIFF
--- a/mathics_django/web/media/js/graphics3d.js
+++ b/mathics_django/web/media/js/graphics3d.js
@@ -2,15 +2,22 @@ const drawFunctions = {
 	point: (element) => {
 		const geometry = new THREE.Geometry();
 
-		geometry.vertices = element.coords.map(
-			(coordinate) => new THREE.Vector3(...coordinate[0])
-		);
+		element.coords.forEach((coordinate) => {
+			const sphere = new THREE.Mesh(
+				new THREE.SphereGeometry(element.pointSize / 2, 24, 24)
+			);
 
-		return new THREE.ParticleSystem(
+			sphere.position.set(...coordinate[0]);
+
+			THREE.GeometryUtils.merge(geometry, sphere);
+		});
+
+		geometry.computeFaceNormals();
+
+		return new THREE.Mesh(
 			geometry,
-			new THREE.ParticleBasicMaterial({
-				color: new THREE.Color().setRGB(...element.color).getHex(),
-				size: element.pointSize
+			new THREE.MeshBasicMaterial({
+				color: new THREE.Color().setRGB(...element.color).getHex()
 			})
 		);
 	},
--- a/mathics_django/web/media/js/graphics3d.js
+++ b/mathics_django/web/media/js/graphics3d.js
@@ -2,24 +2,24 @@ const drawFunctions = {
 	point: (element) => {
 		const geometry = new THREE.Geometry();
 
-		element.coords.forEach((coordinate) => {
-			geometry.vertices.push(new THREE.Vector3(...coordinate[0]));
-		});
+		geometry.vertices = element.coords.map(
+			(coordinate) => new THREE.Vector3(...coordinate[0])
+		);
 
 		return new THREE.ParticleSystem(
 			geometry,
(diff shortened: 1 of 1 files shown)


TiagoCavalcanteTrindade:
@rocky I think that the default `PointSize` (from Mathics core) is too small (it looks smaller than before).

rocky:
> @rocky I think that the default `PointSize` (from Mathics core) is too small (it looks smaller than before).

It always has been very small. The point size is a fraction of the image width which we are not setting properly. In fact, probably for json, we should leave it as a relative size and have the JS code compute its absolute size. What do you think?

TiagoCavalcanteTrindade:
@rocky I think it makes sense as Mathics core can have various front ends each one with it's proper size.

So I should do the size to be what? 400*`PointSize`

rocky:
> @rocky I think that the default `PointSize` (from Mathics core) is too small (it looks smaller than before).

I just checked the Mathematica book (5 edition) and you are correct. I just adjusted the size to 0.01 (for 3D grapics) on master.

rocky:
> @rocky I think it makes sense as Mathics core can have various front ends each one with it's proper size.
> 
> So I should do the size to be what? 400*`PointSize`?

My understanding of how this works is that you should get find out the image dimensions, say the "width" and use that instead of 400.  If there is no image width we've been using 400. 

I will soon change Mathics core so that it doesn't do this multiplication, but instead passes the value as is set in PointSize.

TiagoCavalcanteTrindade:
@rocky by the width you mean what? The canvas width?

If it be the canvas width it's 600*600 but it has a style of 400*400, what is a bit confusing (the effective size is 400*400).

rocky:
> @rocky by the width you mean what? The canvas width?
> 
> If it be the canvas width it's 600*600 but it has a style of 400*400, what is a bit confusing (the effective size is 400*400).
> 
> Also what will be done by now? Wait Mathics core be updated or merge it now?

I guess use the effective size. (If that's wrong I suspect it is easy to change later.) 

In general there are lots of bugs, and I don't think we'll be able to fix all of them at once so, let's get reasonably close. (Is turning the point into a black circle/sphere easily possible)? 

As for Mathics core, the changes are now in https://github.com/mathics/Mathics/commit/0c34b562c3ff017c94925a7a2f5bae0a282e29d7

But in general, I'd wait on stuff to get into Mathics Core first so we can be sure that this works.

TiagoCavalcanteTrindade:
@rocky ok, I'll wait Mathics core be updated, I'm working in the rounded points.

rocky:
> @rocky ok, I'll wait Mathics core be updated, I'm working in the rounded points.

Great! master of Mathics core _is_ now updated so that you always get the relative pointsize back now, and the default is set to 0.01. Hopefully when this is done Mathics  results will look more like https://reference.wolfram.com/language/ref/PointSize.html

TiagoCavalcanteTrindade:
@rocky I think that the best viable fix is draw spheres instead of points (the other ones are: use a texture, but I think this one is too hacky and use a custom shader, this one is the right way but it's still out of my scope).

I choose 24 width and height segments because of it isn't too hard to render (but still harder than just draw particles) and for big points it doesn't look so bad.

rocky (approved):
LGTM - Thanks!

Fixes #62

TiagoCavalcanteTrindade:
@rocky shouldn't #62 be closed?

rocky:
> @rocky I think that the best viable fix is draw spheres instead of points (the other ones are: use a texture, but I think this one is too hacky and use a custom shader, this one is the right way but it's still out of my scope).
> 
> I choose 24 width and height segments because of it isn't too hard to render (but still harder than just draw particles) and for big points it doesn't look so bad.

Ok - make sense. Thanks for handling. Maybe in a future three.js this is addressed in a simpler manner.

BTW https://github.com/Mathics3/mathics-omnibus/commit/2642f69fb60ab08699b4acce51cc9716fa365e55 now includes this in the 3D Graphics Test cases.

PR #64, opened 2021-06-08
Update three.js to r67

TiagoCavalcanteTrindade:
three.js r68 has an error which is only corrected in three.js r69 and update three.js to r69 leads to a bug in `Plot3D` function (but not in `Graphics3D` 🤔).


TiagoCavalcanteTrindade:
Also @rocky I think we should delete some of the old merged branches.

rocky (approved):
Looks good. Made one small change so that /about picks up the revision number in its display.

I tried looking on the 3D Graphics Test Suite gallery, and 14: Coordinates & Graphics 3D and it all looks good. 

Many thanks for getting us to 2014!

rocky:
> Also, @rocky I think we should delete some of the old merged branches.

Ok - a number of the older and less-important ones have been removed now.

PR #65, opened 2021-06-08
Update three.js to r124

TiagoCavalcanteTrindade:
This update three.js to r124 (three.js had almost no changes in the last versions).

As three.js r124 doesn't support canvas rendering I deleted all code associated with it (also the Detector.js file).

This also improve performance of drawing multiple spheres a time.

And revert the behavior change of drawing points introduced in #64 (points increase size when zoomed in) at the same time that increase the performance of drawing multiple points (I converted spheres to a custom GLSL fragment shaper).

The size of the points is the almost same than in r67 (the change is just about 1 pixel of difference in my tests, but I preferred to keep the multiplier 84 than adding a crazier magic number, but if you think it isn't the best feel free to tell me).

This also replace "0." by "." in the axes numbers.


rocky:
@TiagoCavalcanteTrindade  Thanks for doing this. I am looking forward to trying it out. However for the next 24 hours or so, I will be very busy. 

As for the default point size, I agree that what is there now is too small.  When I look at 
```
Graphics3D[Table[Point[{x, y, z}], {x, 5}, {y, 5}, {z, 5}]]
```
from "14: Coordinates and Graphics 3D"  adapted from https://www.wolfram.com/language/elementary-introduction/2nd-ed/14-coordinates-and-graphics.html 

what we have is too small to see. 

Lastly, I will mention that I am not happy with the current behavior of rendering 3D Graphics inside MathML   which doesn't do us any good,  except get in the way.

On the Mathics Core side, I'll be adding a JSONForm kind of format which basically just strips out the MathML header part and leaves the JSON. Then on the Django side, there won't be MathML for 3D objects. 

I supposed we could also do this for 2D objects as well if threejs works well with this or can handle SVG (which is really how 2D is handled). 

If we strip away the MathML, does three.js give us more capabilities in terns of user interaction? 

(Also, I am assuming this doesn't address the polygon issue. 

And there is another Polygon aspect - independent of the 3D aspect, WL polygons use an even/odd rule rather than the convex region for deciding what is in a polygoin. See the Django doc for Polygon and try 

```
Graphics3D[Polygon[{{150,0,0},{121,90,0},{198,35,0},{102,35,0},{179,90,0}}]]
```

Thanks for all of the elp.

TiagoCavalcanteTrindade:
> (Also, I am assuming this doesn't address the polygon issue.

@rocky no it's still there.

> `Graphics3D[Polygon[{{150,0,0},{121,90,0},{198,35,0},{102,35,0},{179,90,0}}]]`

I guess it should draw a triangle, isn't it?

Also about the z axis I'm looking for a solution, it's hard to draw because of it can't be done with just two faces.

rocky:
> 
> 
> > (Also, I am assuming this doesn't address the polygon issue.
> 
> @rocky no it's still there.
> 
> > `Graphics3D[Polygon[{{150,0,0},{121,90,0},{198,35,0},{102,35,0},{179,90,0}}]]`
> 
> I guess it should draw a triangle, isn't it?
> 

It should draw a 5 sided star with the center white (or outside). 

> Also about the z axis I'm looking for a solution, it's hard to draw because of it can't be done with just two faces.

Ok. Thanks for the information.

> 
> Also, if it's too small I guess the multiplier should be 400 (the size of the canvas). I just set 84 because of it was closer to the actual size than 85 or 83.


> 
> > On the Mathics Core side, I'll be adding a JSONForm kind of format which basically just strips out the MathML header part and leaves the JSON. Then on the Django side, there won't be MathML for 3D objects.
> > I imagine we could also do this for 2D objects as well if threejs works well with this or can handle SVG (which is really how 2D is handled).
> 
> I think it's the right way to do the things, as data should be used in any front end, even one which doesn't support MathML

Yep. 
> 
> > If we strip away the MathML, does three.js give us more capabilities in terns of user interaction?
> 
> What type of user interaction? I couldn't in any 😄.


I was just imagining that three.js has these ways to add sliders an zoom bars. It looks like that happens in the three.js demos. And there is mention of "editors". How does one get access to all of that? 

As for MathJax if you right click it gives a menu that is useless for three.js graphics. For MathML it will allow you to zoom the formula, but here it gets in the way. 

BTW I added a small commit to track finding the release number in http://localhost:8000/about

And in Firefox I on Windows in the console messages I am seeing: 

```
Error in parsing MathML attribute value ‘3’ as length.  Attribute ignored.
```

when I run the 3D Graphics Test Suite.

Ok - I gotta get back to work.... See you in a day or so.

rocky:
I just realized that the console message is a MathML message so that will disappear when MathML does, I guess.

TiagoCavalcanteTrindade:
> I was just imagining that three.js has these ways to add sliders an zoom bars.

If you mean the following, it should be easy to add. Just add a GUI Helper with the property you want to change and boom.
image
Example of code:
```js
const control  = new dat.GUI({ autoPlace: true });
control.domElement.id = 'gui';
const folder = datGui.addFolder(`Cube`)
folder.addColor(new ColorGUIHelper(material.cube,'color') ,'value')
    .name('color')
    .onChange(animate)
```

---

> And there is mention of "editors". How does one get access to all of that?

If you mean the following, it's VERY hard:
newVersion

rocky:
> Also about the z axis I'm looking for a solution, it's hard to draw because of it can't be done with just two faces.
> 

@TiagoCavalcanteTrindade  I just looked again at https://reference.wolfram.com/language/ref/Polygon.html and read: 

As a geometric region, the points pi can have any embedding dimension, but must all lie in a plane and have the same embedding dimension.

TiagoCavalcanteTrindade:
@rocky it'll be possible to do it but I think it's better we create a polygon drawing better than the one from Mathematica, but if I couldn't do it in a few days we could add an check to see which 2 coordinates we should use (e.g. if z changes but x is always 0.)

rocky (approved):
Looks great! Thanks for getting us up into the modern era of Three.js.

mathics_django/web/media/js/graphics3d.js
```diff
@@ -2,26 +2,28 @@ const drawFunctions = {
...
-				color: new THREE.Color().setRGB(...element.color).getHex()
-			})
+		geometry.vertices = element.coords.map(
+			(coordinate) => new THREE.Vector3(...coordinate[0])
 		);
+
+		return new THREE.Points(geometry, new THREE.ShaderMaterial({
+			transparent: true,
+			uniforms: {
+				// 84 is the scale for this size of canvas
```
rocky, line 12: I was playing around with this value today with: 

```
Graphics3D[Table[Point[{x, y, z}], {x, 5}, {y, 5}, {z, 5}]]
```

and 200 seems to be a reasonable number. 200 is also used in toCanvasCoords() on line 552. 

Ideally though the magic number 200 would be set as a global constant. Or even better it is probably should be adjustable and set as an attribute inside some object either global, or found in one of the passed-in parameters.

TiagoCavalcanteTrindade: @mmatera [here](https://github.com/Mathics3/mathics-django/issues/66) has the fix of the size of the canvas (it's relative size, I just need to add a multiplier of 0.5, I didn't see the comment of @rocky before).

rocky: > Hi,
> I was trying to follow this discussion, and maybe I haven't done it completely.

It goes without saying (but I will say it anyway), that it is always good to hear from you!

> 
> ...
> The second issue maybe could be fixed if we can pass a relative size for the dot size instead of an absolute one. Would be that possible in three.js?

This was done in https://github.com/mathics/Mathics/pull/1411

and I renamed the less helpful name `PointSize.get_size()` to `PointSize.get_absolute_size()` later on.

In https://github.com/mathics/Mathics/pull/1415 I tried changing asymptote to use circles for points which are adjustable in the circle diameter while the dot size as far as I know isn't. 

However there is something incomplete and this didn't pass CircleCI's testing in using asymptote to build the doc. So I have these changes in comments and at some point we will come back to it.

In general though, I think using a relative size in the specification of PointSize and leaving at that for the formatter is the right way to go, and the extra code and complexity which was in the code base right now is not helpful.

PR #67, opened 2021-06-10
Detangle MathML from 3d graphics

rocky:
We need to merge in the corresponding code in Mathics. But after that's done this should be ready to go.

I plan on releasing a new version of Mathics and Mathics-django after all of this clears.

Fixes #66.

PR #68, opened 2021-06-10
Add style guide

TiagoCavalcanteTrindade:
This add a `.editorconfig` file.

@rocky and @mmatera could you review the style guide?

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,8 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = false
+charset = utf-8
+indent_style = tab
+indent_size = 4 


rocky:
@TiagoCavalcanteTrindade Can we please merge in the other direction. 2c18f03 is completely independent of the others and could go in by itself. 

Removing scriptaculous could go in by itself (or you can combine it with 2c18f03).

Reformatting files _without_ improving them can go in by itself. 


After this is done, #67 would be rebased and it would be a lot smaller and more germane to what it is trying to do.

TiagoCavalcanteTrindade:
> Removing scriptaculous could go in by itself (or you can combine it with 2c18f03).

I didn't remove it, just some files of its folder.

@rocky you mean create a branch for removing unnecessary files and another for reformatting?

About the unnecessary files I think it makes sense, about the reformatting I guess it should be here.

After we finish the `.editorconfig` I'll reformate all files (doesn't makes sense have a style guide and doesn't respect it).

TiagoCavalcanteTrindade:
@rocky I removed the bloat, now this just add the `.editorconfig` file.

PR #71, opened 2021-06-10
Remove unnecessary files from the folder scriptaculous

TiagoCavalcanteTrindade:
This branch remove the following files:
- builder.js (imported but never used)
- scriptaculous.js (never imported)
- unittest.js (never imported)


rocky (approved):
Again thanks for noticing and addressing.

PR #74, opened 2021-06-12
Mathml removal in 2d graphics

rocky:
Render Graphics as SVG without the need for MathML embedding. Similar to what we did for 3D Graphics.

This needs the https://github.com/mathics/Mathics/pull/1419 to work.

### DIFF
--- a/mathics_django/web/format.py
+++ b/mathics_django/web/format.py
@@ -22,6 +22,21 @@ def format_output(obj, expr, format=None):
     evaluation.py format_output() from which this was derived is similar but
     it can't make use of a front-ends specific capabilities.
     """
+
+    def eval_boxes(result, fn, obj):
+        try:
+            boxes = fn(evaluation=obj)
+        except BoxError:
+            boxes = None
+            if not hasattr(obj, "seen_box_error"):
+                obj.seen_box_error = True
+                obj.message(
+                    "General", "notboxes", Expression("FullForm", result).evaluate(obj)
+                )
+
+        return boxes
+
+
     if format is None:
         format = obj.format
 
@@ -49,7 +64,6 @@ def format_output(obj, expr, format=None):
         result = Expression("StandardForm", expr).format(obj, "System`MathMLForm")
 
     # This part was derived from and the same as evaluation.py format_output.
-
     if format == "text":
         result = expr.format(obj, "System`OutputForm")
     elif format == "xml":
@@ -65,21 +79,16 @@ def format_output(obj, expr, format=None):
             result = expr.format(obj, "System`OutputForm")
         elif head == 'System`String':
             result = expr.format(obj, "System`InputForm")
+        elif head == 'System`Graphics3D':
+            form_expr = Expression("StandardForm", expr)
+            result = form_expr.format(obj, "System`StandardForm")
+            return eval_boxes(result, result.boxes_to_js, obj)
         else:
             result = Expression("StandardForm", expr).format(obj, "System`MathMLForm")
     else:
         raise ValueError
 
-    try:
-        boxes = result.boxes_to_text(evaluation=obj)
-    except BoxError:
-        boxes = None
-        if not hasattr(obj, "seen_box_error"):
-            obj.seen_box_error = True
-            obj.message(
-                "General", "notboxes", Expression("FullForm", result).evaluate(obj)
-            )
-    return boxes
+    return eval_boxes(result, result.boxes_to_text, obj)
 
 
 cached_pair = None
--- a/mathics_django/web/format.py
+++ b/mathics_django/web/format.py
@@ -15,6 +15,7 @@ FORM_TO_FORMAT = {
     "System`OutputForm": "text",
 }
 
+
 def format_output(obj, expr, format=None):
     """
     Handle unformatted output using the *specific* capabilities of mathics-django.
@@ -35,7 +36,6 @@ def format_output(obj, expr, format=None):
                 )
 
         return boxes
-
 
     if format is None:
         format = obj.format
@@ -75,11 +75,11 @@ def format_output(obj, expr, format=None):
         if str(expr) == "-Graph-" and hasattr(expr, "G"):
             return format_graph(expr.G)
         head = str(expr.get_head())
-        if head == 'System`CompiledFunction':
+        if head == "System`CompiledFunction":
             result = expr.format(obj, "System`OutputForm")
-        elif head == 'System`String':
+        elif head == "System`String":
             result = expr.format(obj, "System`InputForm")
-        elif head == 'System`Graphics3D':
+        elif head == "System`Graphics3D":
             form_expr = Expression("StandardForm", expr)
             result = form_expr.format(obj, "System`StandardForm")
             return eval_boxes(result, result.boxes_to_js, obj)
--- a/mathics_django/web/format.py
+++ b/mathics_django/web/format.py
@@ -24,9 +24,9 @@ def format_output(obj, expr, format=None):
     it can't make use of a front-ends specific capabilities.
     """
 
-    def eval_boxes(result, fn, obj):
+    def eval_boxes(result, fn, obj, **options):
         try:
-            boxes = fn(evaluation=obj)
+            boxes = fn(evaluation=obj, **options)
         except BoxError:
             boxes = None
             if not hasattr(obj, "seen_box_error"):
@@ -83,6 +83,10 @@ def format_output(obj, expr, format=None):
             form_expr = Expression("StandardForm", expr)
             result = form_expr.format(obj, "System`StandardForm")
             return eval_boxes(result, result.boxes_to_js, obj)
+        elif head == "System`Graphics":
+            form_expr = Expression("StandardForm", expr)
+            result = form_expr.format(obj, "System`StandardForm")
+            return eval_boxes(result, result.boxes_to_svg, obj)
         else:
             result = Expression("StandardForm", expr).format(obj, "System`MathMLForm")
     else:
--- a/mathics_django/web/format.py
+++ b/mathics_django/web/format.py
@@ -92,6 +92,8 @@ def format_output(obj, expr, format=None):
     else:
         raise ValueError
 
+    if result is None:
+        return f"Error in evaluating {expr}"
     return eval_boxes(result, result.boxes_to_text, obj)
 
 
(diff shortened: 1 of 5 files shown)


TiagoCavalcanteTrindade:
@rocky should I test with master of Mathics Core or svg-graphics?

TiagoCavalcanteTrindade (approved):
Looks great!

rocky:
> @rocky should I test with master of Mathics Core or svg-graphics?

Only SVG-graphics will work until we merge that to core. And the only reason I have been holding off is that we may want to allow parameters to `to_svg()` to have a way to specify height and width (optionally). My take is to add this as a catch-all optional "graphics_options" dictionary.

TiagoCavalcanteTrindade:
> Only SVG-graphics will work until we merge that to Mathics Core's master branch.

I merged that branch (just to test) and it is working to me.

image
(the axes numbers just don't look so good, but I guess this is because of force dark mode in my browser)

> And the only reason I have been holding off is that we may want to allow parameters to to_svg() to have a way to specify height and width (optionally). My take is to add this as a catch-all optional "graphics_options" dictionary.
> What is your thought?

@rocky shouldn't the size be set in the front end as SVG can scale to any size?

rocky:
> > Only SVG-graphics will work until we merge that to Mathics Core's master branch.
> 
> I merged that branch (just to test) and it is working to me.
> 
> image
> (the axes numbers just don't look so good, but I guess this is because of force dark mode in my browser)
> 
> > And the only reason I have been holding off is that we may want to allow parameters to to_svg() to have a way to specify height and width (optionally). My take is to add this as a catch-all optional "graphics_options" dictionary.
> > What is your thought?
> 
> @rocky shouldn't the size be set in the front end as SVG can scale to any size?

Yes, you are correct the size should be set in the front end so SVG can scale to any size. It needs to be passed as a parameter in https://github.com/Mathics3/mathics-django/pull/74/files#diff-675ef71ac1114124713b67f8f2a4cf78669ddced161318554df7ad02d3172349R89 and then passed along to `to_svg()`  and that then means a change to on the Mathics core side which is fine.

I'll make the change on the Mathics core side to allow the parameters. Would you make the improvements on the Mathics Django side to do the rest?

TiagoCavalcanteTrindade:
> Yes, you are correct the size should be set in the front end so SVG can scale to any size. It needs to be passed as a parameter in https://github.com/Mathics3/mathics-django/pull/74/files#diff-675ef71ac1114124713b67f8f2a4cf78669ddced161318554df7ad02d3172349R89 and then passed along to to_svg() and that then means a change to on the Mathics core side which is fine.

Sorry, I didn't understand this, right now it needs to, or it'll need to?

> Would you make the improvements on the Mathics Django side to do the rest?

Yes, I'd like!

rocky:
> > Yes, you are correct the size should be set in the front end so SVG can scale to any size. It needs to be passed as a parameter in https://github.com/Mathics3/mathics-django/pull/74/files#diff-675ef71ac1114124713b67f8f2a4cf78669ddced161318554df7ad02d3172349R89 and then passed along to to_svg() and that then means a change to on the Mathics core side which is fine.
> 
> Sorry, I didn't understand this, right now it needs to, or it'll needs to?

I was waffling on this depending on whether you'd do and how disruptive of the code it is.  A problem I have with the Mathics code base is that if I try to do too much in one go something breaks. So I have been doing changes little by little.

(Think back on your experience in trying to go from three.js r52 to r100+) 

In https://github.com/mathics/Mathics/pull/1419/commits/4595a0a66b7a50fc3cd40c12070673436ffd6090 I just added allowing parameters to get set from the outside. 

I haven't tested this to see if this is adequate. (Right now I don't even know if it will pass tests.)

If it is adequate let me know and we can merge. If not bang on it until it is sufficient. 


> 
> > Would you make the improvements on the Mathics Django side to do the rest?
> 
> Yes, I'd like!

And it is both much appreciate and needed.

TiagoCavalcanteTrindade:
@rocky just modified a bit.

I'd just change 2 things in Mathics Core:
- remove the method `boxes_to_mathml` as now it just returns `-Graphics3D-`
- I don't know if it is easy to do but if it's: remove the new line before the SVG

rocky:
> @rocky just modified a bit.
> 
> I'd just change 2 things in Mathics Core:
> 
> * remove the method `boxes_to_mathml` as now it just returns `-Graphics3D-`

I don't follow.  boxes_to_mathml is used when 3D graphics are embedded inside MathML For example: 

```
Table[Graphics3D[{PointSize[r], Point[{0, 0, 0}]}], {r, {0.05, 0.1, 0.8}}]
```

> * I don't know if it is easy to do but if it's: remove the new line before the SVG

Ok removed in https://github.com/mathics/Mathics/pull/1419/commits/b00e8cb35c30f0c160dfd11fd8ce30cf7f7bb524 which has to be coordinated with  ac382ecb5e127bdf64fb01bf2f605612465711ac 

This is a bigger deal than it should be because the unit test in Mathics core looks for that string (and now this is more generic which might not be an improvement. I don't know. On the other hand I do dislike the fragility of the unit test,

Also, I added the newline so that when dumping data it would be more visually distinct, but that's not worth haggling over. 
I also don't like those "startswith" tests in javascript which are also fragile and it means that when someone want to change the formatting of the svg tag things may break. 

Overall it would be nice to remove or reduce those these hidden dependencies between mathics.js and something in Mathics core and its test. 


> 
> Also, I think me we can merge now.

The process I was hoping for what that you'd merge the 3D PR separately. Then I'll rebase this off of master. History wise, it would be nicer to have a PR for 3D off of master and 2D also off of master. I have the base here on 3D simply because you haven't merged it into master yet.

TiagoCavalcanteTrindade:
@rocky ok.

About the `startsWith` it can be replaced by Regex `test` or `indexOf`.

`indexOf` is much faster than `startsWith` (Regex `test` is slower), just wondering if it's too hacky, as we want to test the string, we don't care to the index of the match. What to you think? (maybe I'm overengineering)

rocky:
> 
> 
> @rocky ok.
> 
> About the `startsWith` it can be replaced by Regex `test` or `indexOf`.
> 
> `indexOf` is much faster than `startsWith` (Regex `test` is slower), just wondering if it's too hacky, as we want to test the string, we don't care to the index of the match. What to you think? (maybe it's premature optimization)
> 

This kind of stuff is a bit over engineering. I was hoping for a better overall mechanism that works with objects rather than strings and more elegant. This whole approach seems kludgy.



> About the `boxes_to_mathml` I just see that it's in the base class right now 😄.

TiagoCavalcanteTrindade:
@rocky I could add graphics 3d as an element of the div, operate over its data attribute (after a JSON parse) and then remove it, no `replace`, `indexOf`, `startsWith` or `test` needed. What do you think?

TiagoCavalcanteTrindade:
@rocky I could add graphics 3d as an element of the div, operate over its data attribute (after a JSON parse) and then remove it. No `replace`, `indexOf`, `startsWith` or `test` needed. What do you think?

I think this approach would work for MathML.

rocky:
> 
> 
> @rocky I could add graphics 3d as an element of the div, operate over its data attribute (after a JSON parse) and then remove it. No `replace`, `indexOf`, `startsWith` or `test` needed. What do you think?
> 
> I think this approach would work for MathML.

Ok. Can the wrapping be done purely inside mathics DJango rather than changing Mathics Core again? If so, and Nike says, jsut do it.

TiagoCavalcanteTrindade:
@rocky this approach can be done without changing Mathics Core.

rocky:
> 
> 
> @rocky this approach can be done without changing Mathics Core.

Ok - then please make the changes both in 3D merge that in to master, and then continue here and merge directly into master  - thanks. 

I'd like to focus on other issues this weekend. But  tomorrow I'll put revise the omnibus docker container with what we have by tomorrow.

Probably next weekend we can do a Mathics core and Django release (and other subsidiary packages that have updates like mathicsscript and mathics-scanner.).

TiagoCavalcanteTrindade:
@rocky now I noticed the following:
```html
<graphics3d data="{"elements": [{"type": "sphere", "coords": [[[0, 0, 0], null]], "radius": 1, "faceColor": [1, 1, 1, 1]}], "axes": {"hasaxes": [false, false, false], "ticks": [[[-1.0, -0.5, 0.0, 0.5, 1.0], [-0.9, -0.8, -0.7000000000000001, -0.6000000000000001, -0.4, -0.30000000000000004, -0.2, -0.1, 0.1, 0.2, 0.30000000000000004, 0.4, 0.6000000000000001, 0.7000000000000001, 0.8, 0.9], ["-1", "-0.5", "0", "0.5", "1"]], [[-1.0, -0.5, 0.0, 0.5, 1.0], [-0.9, -0.8, -0.7000000000000001, -0.6000000000000001, -0.4, -0.30000000000000004, -0.2, -0.1, 0.1, 0.2, 0.30000000000000004, 0.4, 0.6000000000000001, 0.7000000000000001, 0.8, 0.9], ["-1", "-0.5", "0", "0.5", "1"]], [[-1.0, -0.5, 0.0, 0.5, 1.0], [-0.9, -0.8, -0.7000000000000001, -0.6000000000000001, -0.4, -0.30000000000000004, -0.2, -0.1, 0.1, 0.2, 0.30000000000000004, 0.4, 0.6000000000000001, 0.7000000000000001, 0.8, 0.9], ["-1", "-0.5", "0", "0.5", "1"]]], "ticks_style": []}, "extent": {"xmin": -1.0, "xmax": 1.0, "ymin": -1.0, "ymax": 1.0, "zmin": -1.0, "zmax": 1.0}, "lighting": [{"type": "Ambient", "color": [0.3, 0.2, 0.4]}, {"type": "Directional", "color": [0.8, 0.0, 0.0], "position": [2.0, 0.0, 2.0]}, {"type": "Directional", "color": [0.0, 0.8, 0.0], "position": [2.0, 2.0, 2.0]}, {"type": "Directional", "color": [0.0, 0.0, 0.8], "position": [0.0, 2.0, 2.0]}], "viewpoint": [2.6, -4.8, 4.0]}" />
```

It's using double quotes everywhere, which will never works. Creating a PR in Mathics Core right now.

rocky:
@TiagoCavalcanteTrindade @mmatera Although both this and the 3D handling have little misfeatures in them and should be removed, I am merging all of this in to so we have a baseline to work from. 

So please keep iterating on 2D and 3D graphics without needing MathML embedding. 

As is my custom, I'll be revising the docker image for what we currently have.

PR #76, opened 2021-06-14
Include quotes when top-level output is a string

rocky:
A small change that I've been thinking about.  I am interested in your thoughts though....

In Mathics 2.0 or so strings were formatted in MathML mode which put them in Sans Serif (usually).  

Previously I changed this so this is no longer in MathML mode and so now strings and symbols are visually more distinct. 
The fact that this is not centered was an artifact that I don't care one about one way or another, but I suspect people would prefer to have this centered like everything else. Your thoughts? 

BTW we probably should separate the cell so that "In" and "Out" are more separate. 

I understand that Mathematica strips off the quotes. I find this annoying not only because you can't tell symbols from strings, but also if you cut and paste the output, you can't use it as input without adding back in the quotes.

But again, I am curious in your views as to how to improve this.

### DIFF
--- a/mathics_django/web/format.py
+++ b/mathics_django/web/format.py
@@ -79,6 +79,10 @@ def format_output(obj, expr, format=None):
             result = expr.format(obj, "System`OutputForm")
         elif head == "System`String":
             result = expr.format(obj, "System`InputForm")
+            # FIXME: In mathicsscript we have a setting --wl-strict-wl-ouput
+            # which controls whether to include quotes when the top-level
+            # output is a string.
+            return result.boxes_to_text(result)
         elif head == "System`Graphics3D":
             form_expr = Expression("StandardForm", expr)
             result = form_expr.format(obj, "System`StandardForm")
--- a/mathics_django/web/media/css/styles.css
+++ b/mathics_django/web/media/css/styles.css
@@ -539,11 +539,12 @@ ul.query li.request {
   padding: 0;
 }
 
-textarea.request {
+.request {
   width: 99%;
+  min-height: 20px;
   overflow: hidden;
   border: none;
-  line-height: 17px;
+  line-height: 20px;
   color: navy;
   margin: 0;
   font-family: monospace;
@@ -552,10 +553,7 @@ textarea.request {
 
   /* hide the blue border in Safari */
   outline-style: none;
-  outline-width: 0px;
-
-  /* disable resize grid in Safari */
-  resize: none;
+  outline-width: 0;
 }
 
 li.out {
@@ -624,8 +622,6 @@ li.test .submitbutton span {
   color: white;
   padding: 5px;
   border-radius: 5px;
-  -moz-border-radius: 5px;
-  -webkit-border-radius: 5px;
   width: 10px;
   line-height: 10px;
   text-align: center;
--- a/mathics_django/web/media/css/styles.css
+++ b/mathics_django/web/media/css/styles.css
@@ -420,7 +420,6 @@ a.pdf {
 
 #queries {
   list-style-type: none;
-  margin: 0;
   padding: 0;
   padding-bottom: 50px;
   padding-left: 0;
@@ -539,12 +538,11 @@ ul.query li.request {
   padding: 0;
 }
 
-.request {
-  width: 99%;
-  min-height: 20px;
+textarea.request {
+  width: calc(100% - 40px);
   overflow: hidden;
   border: none;
-  line-height: 20px;
+  line-height: 17px;
   color: navy;
   margin: 0;
   font-family: monospace;
@@ -554,6 +552,8 @@ ul.query li.request {
   /* hide the blue border in Safari */
   outline-style: none;
   outline-width: 0;
+
+  resize: none;
 }
 
 li.out {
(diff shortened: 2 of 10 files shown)


TiagoCavalcanteTrindade:
> Previously I changed this so this is no longer in MathML mode and so now strings and symbols are visually more distinct.

I think they should be.

> The fact that this is not centered was an artifact that I don't care one about one way or another, but I suspect people would prefer to have this centered like everything else. Your thoughts?

I think we shouldn't alienate to Mathematica, but I googled it and Mathematica doesn't center the strings. Also UX isn't my best.

> BTW we probably should separate the cell so that "In" and "Out" are more separate.

Yes, the options we have are:
- add more space, I think this isn't the best idea;
- add a `<hr />`, I think this doesn't looks good;
- instead of a border in the whole cell, a border around the input and other around the output, I think this would be confusing as differently from mathicsscript, Mathics Django doesn't have the `In`s and `Out`s;
- add two borders around the input and output, I think this is too much information for the eyes;
- add a border around the output;
- add a border around the input, this is what Jupyter notebooks do, so I think it's the best.

---

Also, about transfer Mathics Core, I think that if we do it, we should transfer the other repositories too.

TiagoCavalcanteTrindade:
@rocky what do you and @mmatera think?  
(I said a border around the input was the best, but now I think a border around the output is the best.

image

rocky:
My sense of UI is also something not to be trusted. So I recommend (and it is a general recommendation) to follow the way something else does it. In particular follow Jupyter  because even if it isn't optimum from some hypothetical UI designer sense, at least it will be consistent with similar kinds of things and hopefully not confusing for people.

And down the line we'll be moving to Jupyter too. (But like so many other things, best first to get things cleaned up in the environment we currently have.)

rocky:
I should also note that the example you have is not something we'll ever see. Right now the decision is on a _top-level_ entity.

There is the larger issue of handling strings _whereever_ they appear in the output, but that is part of a larger problem and discussion of the whole output formatting revamp and this is its own separate mess. So for now I limit things to just this somewhat special case.

TiagoCavalcanteTrindade:
@rocky here how Jupyter does.

image
(for me this doesn't looks good)

I also added some color to strings and centered them.

I'll push my modifications and then you can test with other queries.

TiagoCavalcanteTrindade:
@rocky I was thinking if we should keep the quotes and then I queried the following:
image

Now I think that we really should keep the quotes, as we have an function which strip them out.

TiagoCavalcanteTrindade:
Also, this fixes the bug mentioned [here](https://github.com/Mathics3/mathics-django/pull/77#issuecomment-860900134).

rocky:
LGTM - but please I'd rather many smaller commits than one combining a number of things together. If something goes wrong it is easier to pinpoint and isolate. 

(I will confess that I and @mmatera and probably others have a habit of piggybacking separate kinds of changes. Let's face it, it is easier to do if you are short on time. And the code needs soooo many changes.)

TiagoCavalcanteTrindade:
@rocky ok, I'll start doing that.

> If something goes wrong it is easier to pinpoint and isolate.

Yes, also much easier to visualizate the diffs, sometimes I exagerate in squashing the things up.

rocky:
So now I learn that with this commit I can't retrieve saved worksheets. 

Try saving http://localhost:8000/#cXVlcmllcz0oKiUyMEFkYXB0ZWQlMjBmcm9tJTNBJTIwaHR0cHMlM0ElMkYlMkZ3d3cud29sZnJhbS5jb20lMkZsYW5ndWFnZSUyRmVsZW1lbnRhcnktaW50cm9kdWN0aW9uJTJGMm5kLWVkJTJGMDEtc3RhcnRpbmctb3V0LWVsZW1lbnRhcnktYXJpdGhtZXRpYy5odG1sJTIwKikmcXVlcmllcz0oKiUyMEFkZCUyMG51bWJlcnMlMjAqKSZxdWVyaWVzPTIlMjAlMkIlMjAyJnF1ZXJpZXM9KColMjBNdWx0aXBseSUyMG51bWJlcnMlMjAqKSZxdWVyaWVzPTEyMzQlMjAlMkIlMjA1Njc4JnF1ZXJpZXM9MTIzNCo1Njc4JnF1ZXJpZXM9MSUyMCUyQiUyMDIlMjAlMkIlMjAzJnF1ZXJpZXM9MSUyMDIlMjAzJTIwNCUyMDUmcXVlcmllcz01JTIwJTVFJTIwMiZxdWVyaWVzPTMlMjAlNUUlMjAoNyUyMDgpJnF1ZXJpZXM9KDQlMjAtJTIwMikqKDMlMjAlMkIlMjA0KSZxdWVyaWVzPSgyOSUyMDEwJTVFMyklMjA3MyZxdWVyaWVzPUFwcGx5JTVCUGx1cyUyQyUyMFJhbmdlJTVCLTMlMkMlMjAzJTVEJTVEJnF1ZXJpZXM9MSUyRjA=

Then see if you can load back in the results.

TiagoCavalcanteTrindade:
@rocky no, also isn't working to me. I'll look at that.

rocky:
If there is a better way to save and load worksheets that is fine too. All of this is real hacky. Useful (if you work around the hackiness), but still hacky.

rocky:
Another thing I an now noticing is that the the output box constrains MathJaX's zoom so that it now pops up with horizontal and vertical scrollbars for a too small window.

TiagoCavalcanteTrindade:
@rocky I'll also look at that, but today I'm a bit busy. Tomorrow I'll have time.

PR #77, opened 2021-06-14
Start removing Prototype and Scriptaculous

TiagoCavalcanteTrindade:
@rocky it isn't complete but I think is better merge before we do too much changes in master.

### DIFF
--- a/mathics_django/web/media/css/styles.css
+++ b/mathics_django/web/media/css/styles.css
@@ -539,11 +539,12 @@ ul.query li.request {
   padding: 0;
 }
 
-textarea.request {
+.request {
   width: 99%;
+  min-height: 20px;
   overflow: hidden;
   border: none;
-  line-height: 17px;
+  line-height: 20px;
   color: navy;
   margin: 0;
   font-family: monospace;
@@ -552,10 +553,7 @@ textarea.request {
 
   /* hide the blue border in Safari */
   outline-style: none;
-  outline-width: 0px;
-
-  /* disable resize grid in Safari */
-  resize: none;
+  outline-width: 0;
 }
 
 li.out {
@@ -624,8 +622,6 @@ li.test .submitbutton span {
   color: white;
   padding: 5px;
   border-radius: 5px;
-  -moz-border-radius: 5px;
-  -webkit-border-radius: 5px;
   width: 10px;
   line-height: 10px;
   text-align: center;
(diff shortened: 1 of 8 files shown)


rocky (commented):
If this happens the dropdowns stop working. 

Recently, I noticed that even before this, some of the buttons have stopped working. If I run open worksheet (the leftmost icon) and and I enter "Cancel", the menu no longer appears.

Probably we should undo the removals previously done or fix this up in some other way. 

In my opinion right now this kind of thing isn't the as important as improving polygon or getting density plot going.

TiagoCavalcanteTrindade:
@rocky I'll fix these things. I forget to test these ones in last time I modified, but this gonna be easy to fix.

TiagoCavalcanteTrindade:
@rocky now the input area is fixed.

The cancel button I guess was a problem while rebasing (in my computer), now it's also fixed.

---

> In my opinion right now this kind of thing isn't the as important as improving polygon or getting density plot going.

I also think now, but when I started doing it there were less things to do. This is why I said to merge before it be finished, because of it'll take time to I work in it again.

rocky:
> @rocky now the input area is fixed.
>

When I tried the input area is now broken and I can't enter input. The "=" on the right-hand side is in a different font and doesn't work. And Ctrl-enter (the alternate way of entering input) also doesn't evaluate the text that is on input.

TiagoCavalcanteTrindade:
@rocky fixing it now.

TiagoCavalcanteTrindade:
@rocky I reverted to an older version of my code.

I already tested the following:
- open, delete and save worksheets;
- global keys (Ctrl+D, Ctrl+O, Ctrl+S);
- popup's yes and no buttons (and its keyboard versions);
- documentation;

rocky:
@TiagoCavalcanteTrindade  I tested everything and things look good. There even seems to be a slight improvement in the text area in being able to resize that. 

Believe me, I know how frustrating it is to work with this code and trying to improve it only to find there is an additional mess that needs attending. And then having to back off on more ambitious changes because of one of these newly discovered "gotcha"s that prevent progress. 

We just came across one in the way density plots are handled. 

I am optimistic that eventually all of this will be sorted out.

rocky:
There is a slight misfeature in that if you try to save a worksheet without giving it a name it appears to "save" it and then blanks out the save and cancel button. I don't think this happened before.

However, nothing appears to be saved. And this is not such a big deal that it needs to be fixed with any urgency. I'm just noting. (We should probably open an issue, but it is just a low priority one for me.)

PR #78, opened 2021-06-16
Fix queries not showing

TiagoCavalcanteTrindade:
@rocky it wasn't an error in saving the worksheet it was an error in opening it 😄.

About rewriting it, what are your ideas? Maybe remove the `SaveForm` and have just JSON?

Before we merge I'll just rebase it.

### DIFF
--- a/mathics_django/web/media/js/inout.js
+++ b/mathics_django/web/media/js/inout.js
@@ -21,13 +21,13 @@ function openWorksheet(name) {
 	});
 }
 
-function showWorksheets(isOpen) {
+function showWorksheets() {
 	new Ajax.Request('/ajax/getworksheets/', {
 		method: 'get',
 		onSuccess: (transport) => {
 			var response = JSON.parse(transport.responseText);
 			var tbody = document.getElementById('openFilelist');
-			tbody.deleteChildNodes();
+			tbody.innerHTML = '';
 			response.worksheets.forEach((worksheet) => {
 				tbody.appendChild($E('tr',
 					$E('td',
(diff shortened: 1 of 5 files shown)


rocky:
> @rocky it wasn't an error in saving the worksheet it was an error in opening it .
> 
> About rewriting it, what are your ideas? Maybe remove the `SaveForm` and have just JSON?

LGTM - Thanks for fixing this. I don't have any ideas for how to redo, other than mention it and, offer the possibility were this too onerous to fix.  If it is fixed let's not contemplate redoing it. It will all take care of itself when we can move stuff to Jupyter. 

As for using JSON more pervasively, I am generally for that. 

Recently, I got the CLI version mathicsscript running so that it can display SVG's via matplotlib. See https://github.com/Mathics3/mathicsscript/pull/40 . Also I've mentioned the great work you are doing on https://groups.google.com/g/mathics-users/c/vxTjgYOg8tQ 

But in untangling all the graphics, I realize that things would have been much simpler, scalable and cleaner if a JSON format were defined for 2D and the code to convert that to SVG, or whatever moved out of core. All of this will make it easier to support Plotly or ds3 or whatever your favorite thing is.

rocky:
About the mathml zoom problem  - the zoom menu seems to allow up to 4 times zooming (400%).

I wonder if there is a way to have things look the same but underneath the boxes overlap in white margins or borders, but since the overall box is large the zoom will use all of that. Or maybe the mathml zoom box can just be set separately.  I really don't know this stuff, so I really shouldn't offer any thoughts about it.

TiagoCavalcanteTrindade:
@rocky about the scrollbars I'll fix them tomorrow.

rocky:
@TiagoCavalcanteTrindade you are doing great work. No rush. 

In general we should think about solidifying things so that we can come out with a new Mathics core, django, and script releases.

So no more major refactoring for now. Let's leave that for after release. (Which no doubt will be another major upheaval)

PR #80, opened 2021-06-16
Remove flickering

### LINKED ISSUE #54: Cell Output Area flicker in (re)evaluation
Since upgrading to the new MathJax, we get lots of re-evaluations in cell data and this causes lots of flickering in the cell area of the display.

Part of this may be caused by the fact that we evaluate twice in some circumstances, I think: once because of the stupid format done automatically, and once done when we realize that's not the format we want.

Part of this may be caused by MathJax's preview mode.

TiagoCavalcanteTrindade:
Fixes #54.

This branch also fixes the lack of scrollbar for very long formulas, e.g.: `Table[x, {x, -20, 20}]`.

### DIFF
--- a/mathics_django/web/media/css/styles.css
+++ b/mathics_django/web/media/css/styles.css
@@ -567,7 +567,7 @@ ul.out {
   border: 1px solid #666;
   list-style-type: none;
   margin: 10px 40px 0 0;
-  overflow: hidden;
+  overflow: auto;
   box-sizing: border-box;
   padding: 0;
   font-family: monospace;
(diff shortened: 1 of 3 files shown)


TiagoCavalcanteTrindade:
@rocky I don't know why it have failed, so I'm re-running the tests.

rocky:
LGTM - thanks!

It also seems to fix problems we had in "07: Colors and Styles"

(PIe charts are still broken in 04: Displaying lists, but that's a separate issue.)

PR #82, opened 2021-06-17
Improve polygon drawing

TiagoCavalcanteTrindade:
This branch implements non triangular 3d polygon drawing.

This branch increase the support for coplanar polygons (aka shapes).

This branch still doesn't implements coplanar polygons with holes.


rocky:
@TiagoCavalcanteTrindade Thanks for undertaking this.  I just tried 

```
Graphics3D[Polygon[{{150,0, 0},{121,90, 0},{198,35, 0},{102,35, 0},{179,90, 0}}]]
```

to see if there is a change here. Not yet. 

In other news, it looks like I botched PieChart: when pulling out the formatting routines, I missed the formatting routine for ArcBox.  (I still have a bit more work to do there as well)

TiagoCavalcanteTrindade:
> @TiagoCavalcanteTrindade Thanks for undertaking this. I just tried
> 
> ```
> Graphics3D[Polygon[{{150,0, 0},{121,90, 0},{198,35, 0},{102,35, 0},{179,90, 0}}]]
> ```
> 
> to see if there is a change here. Not yet.
> 
> In other news, it looks like I botched PieChart: when pulling out the formatting routines, I missed the formatting routine for ArcBox. (I still have a bit more work to do there as well)

No, to do it we need to calculate the holes.

rocky:
> 
> No, to do it we need to calculate the holes.

I wasn't thinking of the holes, but of the overall star shape. I took the 2D  5-sided star and added a z component with 0. 


> 
> Here are some of the improvments:
> 
> ```mathematica
> Graphics3D[
>     Polygon[{
>         {0, 0, Sqrt[3]},
>         {0, 2 Sqrt[2/3], 1 / Sqrt[3]},
>         {Sqrt[2], Sqrt[2/3], -1 / Sqrt[3]},
>         {Sqrt[2], -Sqrt[2/3], 1 /  Sqrt[3]},
>         {0, -2 Sqrt[2/3], -1 / Sqrt[3]},
>         {-Sqrt[2], -Sqrt[2/3], 1/Sqrt[3]},
>         {-Sqrt[2], Sqrt[2/3], -1/Sqrt[3]}
>     }]
> ]
> ```
> 
> image
> 
> ```mathematica
> Graphics3D[Polygon[{{1,0,0}, {1,1,1}, {1,1,0}}], Axes->True]
> ```
> 
> image
> 
> @rocky if the above are wrong could you tell me? (I'm not that good of drawing polygons in my mind )

Thanks. I'll look later and maybe we can add these to the gallery and examples.

There is a bit that seems intermittently broken in the examples. From "07: Colorfs and Styles":

```
Red
```

```
{Red, Green, Blue, Purple, Orange, Black}
```

From 04: Displaying Lists: 

```
{BarChart[{1, 1, 4, 2}], BarChart[{5, 1, 1, 0}], 
 BarChart[{1, 3, 2, 4}]}
```

TiagoCavalcanteTrindade:
> I wasn't thinking of the holes, but of the overall star shape. I took the 2D 5-sided star and added a z component with 0.

@rocky for me it has a hole in 2d.

image

rocky:
> > I wasn't thinking of the holes, but of the overall star shape. I took the 2D 5-sided star and added a z component with 0.
> 
> @rocky for me it has a hole in 2d.
> 
> image

Yes, it has a hole, but I would have thought that a better polygon routine would have figured out the star-ness  whether or not there was a hole in it.

TiagoCavalcanteTrindade:
@rocky yes, this is a better routine, I'll implement it. I'll look at the Mathics Core code that do it for 2d then translate to JavaScript.

TiagoCavalcanteTrindade:
@rocky I'm searching by the algorithm that calculate where the holes are, but I didn't find it, do you know where is it?
I've searched in `Polygon`, `PolygonBox`, `_Polyline`, `GraphicsBox` (a big class but it doesn't look to be here) and `BoxConstruct`

rocky:
It

> @rocky and @mmatera I'm searching by the algorithm that calculate where the holes are, but I didn't find it, do you know where is it?
> I've searched in `Polygon`, `PolygonBox`, `_Polyline`, `GraphicsBox` (a big class but it doesn't look to be here) and `BoxConstruct`.

It is not in Mathics code, but done by SVG natively using the fill rule "evenodd" in `svg.py`.  

I looked how the SVG implementation "inkscape" does this in C++ and it is pretty long and involved: https://gitlab.com/inkscape/inkscape/-/blob/master/src/path/path-boolop.cpp#L334-850

I guess we just leave this for later then and go with something like this.

TiagoCavalcanteTrindade:
@rocky do you want I implement the [algorithm which checks if the shape has a hole](https://en.wikipedia.org/wiki/Even%E2%80%93odd_rule)?

I was wondering if it makes sense, as we won't implement polygons with holes, and polygons without holes are being drawn correctly.

rocky:
> @rocky do you want I implement the [algorithm which checks if the shape has a hole](https://en.wikipedia.org/wiki/Even%E2%80%93odd_rule)?
> 
> I was wondering if it makes sense, as we won't implement polygons with holes, and polygons without holes are being drawn correctly.

It looks like the thing that would be the most useful here is [CrossingPolygon](https://reference.wolfram.com/language/ref/CrossingPolygon.html) 

Also, since right now we are not handling polygons without holes properly and this PR addresses that better. 

Lastly, I'll mention that there is a big cleanup effort I'd like to see before going off and doing too much additional stuff.

@mmatera  Right now in Mathics core I am planning on separating the boxing functions into its own directory rather than have that embedding inside each module for a builtin. The builtin module will just import the related boxing routine. So maybe  
`builtin/graphics.py` `builtin/box/graphics.py`, `builtin/drawing/plot.py`, `bultin/box/drawing/plot.py`. I am not totally sure on this.

Also I'd pull out Form bultins and put that in its own module: `builtin/form/mathml`, `builtin/form/tex`, etc.

The `formatter` directory will get renamed to `format` (to match the nouns `builtin`, `box`, and `form`).  I'd like more stuff inside JSON so that we have JSON more JSON representations of higher-level object that front-end that have access to threejs, or matplotlib or plotly can use.  And finally there is the issue, that depending on a front-end's capabilities it may prefer to use its own implementation of various higher level items. So there should be a way to tell Mathics, that we'd like our own implementation of Pie chart (a high-level thing) or our own implementation of ArrowLine (a somewhat lower level thing). 

But all of this is open for review. So your thoughts?

TiagoCavalcanteTrindade:
> Lastly, I'll mention that there is a big cleanup effort I'd like to see before going off and doing too much additional stuff.

Ok, I'll focus in the bugs.

rocky (approved):
This is an improvement and in the short term, it looks like we're not likely to do better.

TiagoCavalcanteTrindade:
@rocky yes, the even odd is hard to implement, but I'll keep my eyes on it.

TiagoCavalcanteTrindade:
@rocky I think it wasn't the end, my hope is [ShapePath](https://threejs.org/docs/#api/en/extras/core/ShapePath) solve this.

PR #83, opened 2021-06-18
Fix share button

TiagoCavalcanteTrindade:
With the `improve-polygon-drawing` branch the share button stopped working, so this branch fix it by removing old code from `createLink` function.  
(this branch doesn't need `improve-polygon-drawing` branch to work)


rocky:
@TiagoCavalcanteTrindade You are the expert here. So do whatever you think best.

PR #87, opened 2021-06-19
Remove link underline between logo and Mathics

rocky:
Having the log and Mathics text share an <a> tag with a space in between
causes there to be a link underline in between them.
I split these into two <a> tags.

Perhaps there is a better way?

### DIFF
--- a/mathics_django/web/templates/about.html
+++ b/mathics_django/web/templates/about.html
@@ -8,7 +8,7 @@
 
 {% block html_body %}
 	<header>
-		<a href="https://mathics.org">
+		<a href="https://mathics.org" style="text-decoration: none">
 			<img id="logo" class="load" src="{% static 'img/logo-heptatom.svg' %}" height="32" alt="Logo" />
 			<img id="logotext" class="load" src="{% static 'img/logo-text.svg' %}" height="26" alt="Mathics" />
 		</a>
--- a/mathics_django/web/templates/base.html
+++ b/mathics_django/web/templates/base.html
@@ -11,7 +11,7 @@
 {% block html_body %}
 	<header>
 		<div id="headerleft">
-			<a href="https://mathics.org">
+			<a href="https://mathics.org" style="text-decoration: none">
 				<img id="logo" class="load" src="{% static 'img/logo-heptatom.svg' %}" height="32" alt="Logo" />
 				<img id="logotext" class="load" src="{% static 'img/logo-text.svg' %}" height="26" alt="Mathics" />
 			</a>


TiagoCavalcanteTrindade:
@rocky line breaks also count as whitespaces, and if we let all children of the link inline it'd better to add a comment explaining why, so I think a `style="text-decoration: none"` would be better.

rocky:
> @rocky line breaks also count as whitespaces, and if we let all children of the link inline it'd better to add a comment explaining why, so I think a `style="text-decoration: none"` would be better.

Ok - thanks for the explanation and fix. Feel free to merge when this clears.

rocky:
By the way, I just noticed this is also showing on the main page in firefox too: 
stray_underline

rocky:
(I never could figure out how to make both the main page and "about" show more element.)

TiagoCavalcanteTrindade:
@rocky we could create a component for the link, but I guess this doesn't worth.  
I was thinking these days in split the backend and frontend of Mathics Django (then we could have the backend of a frontend and its frontend too 😆). This would be faster, and the code would be more uncoupled, and also would be easier in the future to replace Django by other framework.

I was thinking that even if a Mathics Jupyter be created Mathics Django shouldn't die, as [at least for me] Jupyter's notebooks never were that fast (not even close to what a Apache/Nginx + Python framework could be).

rocky:
> @rocky we could create a component for the link, but I guess this doesn't worth.
> I was thinking these days in split the backend and frontend of Mathics Django (then we could have the backend of a frontend and its frontend too ). This would be faster, and the code would be more uncoupled, and also would be easier in the future to replace Django by other framework.

If you think it is better - go for it!

> 
> I was thinking that even if a Mathics Jupyter be created Mathics Django shouldn't die, as [at least for me] Jupyter's notebooks never were that fast (not even close to what a Apache/Nginx + Python framework could be).

The main consideration here was that we never had that many people, so he have to limit what we do based on the people we have available and previously we never had anyone interested in Mathics Django. 

Right now Mathics Django has been very helpful in helping us understand front-end API issues and issues in general.  If you want to replace it with something else like flask - go for it.  The intention was get things organized in away to _allow_ the possibility for multiple front ends (not limit them which is what the pre 2.0.0 monolithic code base was doing).

I have just invited you to the Mathics3 team. If you want to start a _new_ project for flask, after accepting the invitation, you should be able to create a new repository under Mathics3.

TiagoCavalcanteTrindade:
@rocky maybe in the future I create a Mathics *framework*, Flask was just a example, with almost sure there's other consolidated and faster framework. But by now I guess the performance improvement of Apache/Nginx would be greater.

Also when I finish pymathics-ml (or whatever we call it), Mathics3 would be a good place to put it.

PR #88, opened 2021-06-19
Add support for smaller screens for SVG

TiagoCavalcanteTrindade:
gif


rocky (approved):
@TiagoCavalcanteTrindade  This is pretty cool. (But as you say I hope in the long run we'll be able to arrange things so that the extraction trick isn't needed.)

PR #89, opened 2021-06-19
Remove html comments

TiagoCavalcanteTrindade:
Remove unnecessary HTML comments.

### DIFF
--- a/mathics_django/web/templates/base_html.html
+++ b/mathics_django/web/templates/base_html.html
@@ -52,48 +52,6 @@
 	{% block html_body %}
 	{% endblock %}
 
-	<!--
-		<div style="display: none">
-		<math id="prototype_math" xmlns="http://www.w3.org/1998/Math/MathML">
-		<mstyle id="prototype_mstyle"></mstyle>
-		<mrow id="prototype_mrow"></mrow>
-		<msup id="prototype_msup"><mi>x</mi><mi>y</mi></msup>
-		<msub id="prototype_msub"><mi>x</mi><mi>y</mi></msub>
-		<msubsup id="prototype_msubsup"><mi>x</mi><mi>y</mi><mi>z</mi></msubsup>
-		<mfrac id="prototype_mfrac"><mi>x</mi><mi>y</mi></mfrac>
-		<msqrt id="prototype_msqrt"><mi>y</mi></msqrt>
-		<mo id="prototype_mo">x</mo>
-		<mn id="prototype_mn">1</mn>
-		<ms id="prototype_ms">x</ms>
-		<mi id="prototype_mi">x</mi>
-		<mtext id="prototype_mtext">t</mtext>
-		<mtable id="prototype_mtable">
-		<mtr id="prototype_mtr">
-		<mtd id="prototype_mtd"></mtd>
-		</mtr>
-		</mtable>
-		</math>
-
-		<svg id="prototype_svg" xmlns:svg="http://www.w3.org/2000/svg" xmlns="http://www.w3.org/2000/svg">
-		<g id="prototype_g" />
-		<rect id="prototype_rect" />
-		<circle id="prototype_circle" />
-		<polyline id="prototype_polyline" />
-		<polygon id="prototype_polygon" />
-		<path id="prototype_path" />
-		<ellipse id="prototype_ellipse" />
-		<foreignObject id="prototype_foreignObject" />
-		</svg>
-
-		<div id="prototype_div_html" xmlns="http://www.w3.org/1999/xhtml"></div>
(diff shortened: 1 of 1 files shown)


TiagoCavalcanteTrindade:
Also, @rocky and @mmatera what do you think of creating a Heroku server of Mathics Django?

mmatera:
That sounds interesting, do you think it would be easy to maintain it?

TiagoCavalcanteTrindade:
@mmatera yes. We could [integrate with this repository](https://devcenter.heroku.com/articles/github-integration). It also would be [free](https://www.heroku.com/pricing).

rocky:
> Also, @rocky and @mmatera what do you think of creating a Heroku server of Mathics Django?

Of all of the problems that I would like to see solved this has been very very low on my list. 

If you think the security issue is no big deal, then I suspect you don't know that much about security or Mathics or both. 

https://mathics-development-guide.readthedocs.io/en/latest/running.html#running-mathics-on-your-server and the next section
https://mathics-development-guide.readthedocs.io/en/latest/running.html#running-mathics-on-a-public-webserver discuss this a little bit. 

I didn't write any of it. It is a holdover from the previous folks that did do just this. So this probably is neither up to date nor complete. I haven't checked to see if `ENABLE_FILES_MODULE` works and does the trick. 


> 
> If you want, I can do it. I just think that to do it we should add a option to disable saving and loading worksheets.

Why disable loading a worksheet? 

Some of the information in the "about" page would need to be disabled and possibly the some of the underlying commands, and possibly those that use `psutil`.  Mathicsscript  and mathics allow you to shell out to the OS to run a command. I don't think though in the Django web server we allow that. 

But over time there probably will be more direct access to Python and when that happens then this needs to be considered. I hope/suspect Heroku has figured out some of this already and so there are probably mechanisms in place. I suspect they also have guidelines to follow as well.

TiagoCavalcanteTrindade:
> Why disable loading a worksheet?

The worksheets are saved in a SQLite database which is erased in Heroku something about 1 time a week, this was just an example of behavior, I thought and save the worksheets in local storage would be better.

> If you think the security issue is no big deal, then I suspect you don't know that much about security or Mathics or both.
https://mathics-development-guide.readthedocs.io/en/latest/running.html#running-mathics-on-your-server and the next section
https://mathics-development-guide.readthedocs.io/en/latest/running.html#running-mathics-on-a-public-webserver discuss this a little bit.

I'll give a look at that, maybe something like each user have one session (as by the start we won't have much of them).

> Some of the information in the "about" page would need to be disabled and possibly the some of the underlying commands, and possibly those that use psutil. Mathicsscript and mathics allow you to shell out to the OS to run a command. I don't think though in the Django web server we allow that.

I can try to fix these security issues.

> But over time there probably will be more direct access to Python and when that happens then this needs to be considered. I hope/suspect Heroku has figured out some of this already and so there are probably mechanisms in place. I suspect they also have guidelines to follow as well.

I'll read the terms of use of Heroku.

rocky:
> Also, @rocky and @mmatera what do you think of creating a Heroku server of Mathics Django?

One other thought here since this has come up a bit. People who are new to this project think, as is natural: "gee there are about 10 obvious way we could extend this".  Some of them involving making it more widely available to casual users (like mathics as a service, or mathics in jupyter). Some of them involve getting running some pretty complicated Mathematica packages on Mathics (e.g. FeynCalc, FeynRules, Rubi). 

From my standpoint, I could spend 6 months just making what this code _currently does_ more solid, sensible, and stable: 

* Slowness in Rule expansion
* Slowness in code overall
* Cleaning up the Assigment rule
* Making the code more modular
* Getting a sensible API for frontends
* Fixing bugs we already know about in the existing builtins 
* Handling the graphics subsystem more completely
* Offloading graphics-like backends to external modules/projects
* Adding/filling in those backends for say matplotlib, plotly, etc
* Making existing builtins more closely match the full WL spec

I could spend 6 months just on the above alone. And were more of these done, I'd feel better about expanding the audience to novices and those who know WL very well were more of these done. 

I fear that if we do say a slipshod job offering mathics as a service, or done inside jupyter (where there probably is already services out there), then we are just opening ourselves to massive bug reports and rightful criticism because as we all know, it doesn't take too long using the system to find some sort of bug. 

And looking at the bugs logged in Mathics core it will just keep getting larger and larger.

So again in sum, I'd rather have something smaller (like the size it is right now that is quite huge) that I and others have _control_ over than expand along ways that will just lead to more bugs and draw us away from making things more solid.

TiagoCavalcanteTrindade:
@rocky now I understand, a service right now would dirty the name of Mathics.

rocky (approved):
LGTM

We got so lost in the side discussion that I had forgotten about the _actual_ change!

PR #90, opened 2021-06-19
Make more clear that this program is Django

rocky:
In the future there may be flask, jupyter, and heroku versions or variations.

Start to make clear what we have here.


mmatera (approved):
LGTM

PR #91, opened 2021-06-20
Misc documentation fixes

rocky:
Module location mathics.doc.doc -> mathics.doc.common_doc
Don't add to online doc sections which are empty

### DIFF
--- a/mathics_django/doc/django_doc.py
+++ b/mathics_django/doc/django_doc.py
@@ -20,7 +20,7 @@ from mathics.core.evaluation import Message, Print
 from mathics_django.doc.utils import escape_html, slugify
 from mathics_django.settings import DOC_XML_DATA_PATH
 
-from mathics.doc.doc import (
+from mathics.doc.common_doc import (
     CHAPTER_RE,
     DocElement,
     END_LINE_SENTINAL,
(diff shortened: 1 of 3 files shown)


rocky:
Note: if/when this clears, I'll be merging this. Of course after the fact you can review. I want to get out another docker image this week.

And then hopefully next week we will be in shape for a release of everything.

TiagoCavalcanteTrindade (approved):
LGTM - I'll look at the `RowBox`.

rocky:
> LGTM - I'll look at the `RowBox`.

Do you mean here or in Mathics core?  Or at for which aspect? (Recall as I have been maintaining, there are generally multiple bugs all over the place and it is too difficult for me to try to fix all of them at once). 

I should mention that I fully expect more rounds of refactoring as we start to get all of this under control. Just like we have been doing for formatting (which is still far from complete). 

One thing that I expect will happen in the future is that the boxing routines will change various default parameters like the default image size from normal to small.  This will help in the display of various images that appear in lists.

But all of that is not for this weekend.

TiagoCavalcanteTrindade:
@rocky I said that because of the FIXME that says `RowBox` have documentation, but it isn't appearing.

rocky:
@TiagoCavalcanteTrindade Ok thanks. The simple thought I had was to add a class variable like 

```
class RowBox()
   user_documentation = True # or False
```

and have this take precedence over whether there is "Box" at the end. (If this class variable doesn't exist, then we default to not showing the docstring as we have now.).

PR #92, opened 2021-06-25
Add Settings`* to http://localhost:8000/about

rocky:


TiagoCavalcanteTrindade (commented):
@rocky do you think we should include the usage here?

rocky:
> 
> 
> @rocky do you think we should include the usage here?

Yes - that makes a lot of sense! 

 (If you are interested just push on top. Otherwise I'll add when I have a chance to get around to it.)

Also, I never bothered styling the about because I am so clumsy at this kind of thing .

TiagoCavalcanteTrindade:
> (If you are interested just push on top. Otherwise I'll add when I have a chance to get around to it.)

Yes, today happily I have time.

TiagoCavalcanteTrindade:
@rocky I don't know if it's just in my machine, but `mathics.doc.doc` is named `mathics.doc.common_doc` here.

rocky:
> 
> 
> > (If you are interested just push on top. Otherwise I'll add when I have a chance to get around to it.)
> 
> Yes, today happily I have time.

Thanks! Here is an even slicker idea: for those things that are Booleans turn into a check box to toggle the setting. 

I will also mention that down the line I would like to change the Mathics core library so that when requested for the usage we can return something that can be formatted to something nicer than a string. Perhaps for HTMLForm or TeX form more nicely formatted usage. I'd like to see a SphinxForm or an RsTForm. But this is down the line...

Also one other thing to note is that over the weekend will be the big release of everything. So whatever can't be done by that will just go into the following release.

There is a lot - more of  the same - kinds of things with respect to organizing functions into their corresponding sections. Perhaps I will finally get around to adding the concept of "Subsection" into the document hierarchy. 

Right now we have Part, Chapter, Section, and Entity, but some and in the future many more don't nicely fit into that for example  

```
Overview / Built-in-Functions /  Colors / Named Colors / Green 
^  part    ^ chapter             ^ section  ^ subsection  ^ entity
```

rocky:
> 
> 
> @rocky I don't know if it's just in my machine, but `mathics.doc.doc` is named `mathics.doc.common_doc` here.

`mathics.doc.common_doc` is the correct module name and your mathics-django installation needs to be updated.  Again, the way to see where your Mathics Django documentation lives is to go into python, import the module (mathics-dango) and see where that resides. And if it isn't the local source then remove it. 

This kind of thing - incompatibility as we reorganize - is one of the reasons we need to release. 

The name `mathics.doc.doc` is kind of stupid. And down the line the stuff that handles running tests from the documentation, storing the data for use in Django, or LaTeX, and the final formatting to PDF all should be moved _out_ of Mathics core.

rocky:
What's is still the case is that just about everywhere you poke more than on surface level everything needs to be redone. Sigh.

TiagoCavalcanteTrindade:
> Thanks! Here is an even slicker idea: for those things that are Booleans turn into a check box to toggle the setting.

Yes, this is a great idea!

TiagoCavalcanteTrindade:
@rocky again, my lack of understanding of the Mathics API:
```py
evaluation.evaluate(f"??{setting_name}").to_python()
```

Where `evaluation` is `get_session_evaluation(request.session)` (https://github.com/Mathics3/mathics-django/blob/master/mathics_django/web/views.py#L116).

Isn't working.

rocky:
> @rocky again, my lack of understanding of the Mathics API:
> 
> ```python
> evaluation.evaluate(f"??{setting_name}").to_python()
> ```

```python
setting_usage_expr = evaluation.parse(setting_name + "::usage")
setting_usage_value = setting_usage_expr.evaluate(evaluation)
```
seems to work.  The exact details of _why_ I am not totally certain.

The fact this kind of stuff isn't obvious and descriptions of why this make sense are hard means that the API needs fixing.

(Did I mention that everywhere you probe deeper, things need to be redone?)   

Before 6 months ago, we didn't have the `to_python()` method and you had to call `.get_string_value()` or `get_int_value()`, etc for the right method for the datatype. 

Before 6 months ago, we didn't have `session` either.

TiagoCavalcanteTrindade:
> The fact this kind of stuff isn't obvious and descriptions of why this make sense are hard means that the API needs fixing.
> 
> (Did I mention that everywhere you probe deeper, things need to be redone?)

Yes for both.

> Before 6 months ago, we didn't have the to_python() method and you had to call .get_string_value() or get_int_value(), etc depending on for the datatype.
> 
> Before 6 months ago, we didn't have session either.

Wow, you made a lot of things in this time, I can't think in Mathics without sessions or `to_python`!

TiagoCavalcanteTrindade:
@rocky this still isn't working (clicking the checkbox does nothing), but I thought was better to push to we don't do the same work twice.

TiagoCavalcanteTrindade:
Something suddenly broke `Settings`$QuotedStrings`.

TiagoCavalcanteTrindade:
@rocky strangely, if I change the value of any settings, the about page gives a timeout error (even setting `TIMEOUT` to a big value).

I still can evaluate this in a query, but I can't load the about page, independently if I change the setting value in a query or in the about page.

rocky:
> 
> 
> @rocky strangely, if I change the value of any settings, the about page gives a timeout error (even setting `TIMEOUT` to a big value).

https://github.com/Mathics3/mathics-django/blob/0e2e3890c28d02930719085e863a9ea95c6875f4/mathics_django/web/views.py#L115 was the code I used to prevent this.

> 
> I still can evaluate this in a query, but I can't load the about page, independently if I change the setting value in a query or in the about page.

TiagoCavalcanteTrindade:
@rocky I removed that unintentionally.

TiagoCavalcanteTrindade:
@rocky it's working now.

I think we can merge.

rocky:
> @rocky it's working now. Also, I rebased my commits.
> 
> I think we can merge.

Overall this looks great! Thanks for doing. 

I am seeing this old recurring problem on Firefox (only) where RGB colors or the graphs inside a matrix are not appearing. 
And I don't know if this is related, but in the console I see  `Error in parsing MathML attribute value '3' as length. Attribute ignored.` 

If you have a chance to look at great!, If not I suppose we can skip - just merge when you are ready.

TiagoCavalcanteTrindade:
@rocky I'll look at it, but at least shouldn't be related.

About the console message, I still didn't fixed it, I guess it is a  problem in MathJax.

rocky:
@TiagoCavalcanteTrindade One other very small think I noticed (and is almost not worth mentioning): I notice now looking at the about page that I messed up in the settings file in that some sentences have periods at the end like the one for "QuotedStrings" and most of them don't. And some aren't even sentences like: 

>  If this Boolean variable is set True, MathMLForm uses a Sans-Serif font 

should be: 

>  If this Boolean variable is set True, MathMLForm use a Sans-Serif font .


I  can  fix up the settings.m file after this goes in or commit here.

TiagoCavalcanteTrindade:
@rocky by me you can commit here.

rocky:
> @rocky by me you can commit here.
> 
> Also, @rocky the color matrixes are working in my Firefox.
> 
> image

Not working for me, but let's skip it for now. It might be something with my firefox. Thanks for checking!

PR #94, opened 2021-06-28
First cut at towards getting subsections fully integrated

rocky:
This should now work.  If you look under "Colors" or "Special Functions" you should see the new level of breakout. 

In the future, Large items "Lists" will be broken out into subsections as per Mathematica 5 or so.

Further reorganizing the actual data on the Core side and adding Pymathics is separate problem.
Also tracking this in the printed manual needs to be done on Core side. (Or better a new repository broken off from Core to handle docs)

However the part I most want - a path towards a more hierarchical organization that is more easily navigable is largely done.


TiagoCavalcanteTrindade:
Looks great!

I'll just do the "rename URL to URI" TODO.

rocky:
Searching for subsection builtins like "Red" doesn't work. Will have to address another day.

TiagoCavalcanteTrindade:
> Searching for subsection builtins like "Red" doesn't work. Will have to address another day.

Ok.

rocky:
> Looks great!
> 
> I'll just do the "rename URL to URI" TODO.

@TiagoCavalcanteTrindade  Great!

Actually if you _finish_ this I'd be grateful. (And we'd have at least another person who understands this better.)

Fixing the interactive search that I messed up is probably around here https://github.com/Mathics3/mathics-django/blob/620ac9542dd8e6a89b190a4b7ede4513dfcc5eae/mathics_django/doc/django_doc.py#L150-L157

where we now have these new kinds of sections called "Guide Sections" that are higher-level groupings of related items. A guess is that Searching is not going down these items where it should. 

Thanks in advance!

rocky:
@TiagoCavalcanteTrindade Something else that would be awesome would be to add some unit tests for this code. *Anything* would be better than the nothing we have right now. Again, thanks!

TiagoCavalcanteTrindade:
@rocky I want to do it, but now I'm working in the 3d lines.

image

rocky:
I

> @rocky I want to do it, but now I'm working in the 3d lines.
> 
> image

Ok - np.  The important thing to know is just whether you will do eventually. Then I can focus on other things when I get a chance.

TiagoCavalcanteTrindade:
> Ok - np. The important thing to know is just whether you will do eventually. Then I can focus on other things when I get a chance.

Oh, I want to do it tomorrow.

You can do other things (the ones I can't).

PR #98, opened 2021-07-01
Avoid MathML for Symbols and Integers

rocky:
For your consideration. Reducing the use of MathML where it isn't needed reduces the need for it to want to reformat it on loading. 

The output changes slightly, and there may be additional styling, or not. 

I am curious to know what you think and what your opinion is.

(Also, down the line we need to come up with a better story and design for Format. Or just hide a lot of the ugliness into its own routine somewhere).

### DIFF
--- a/mathics_django/web/format.py
+++ b/mathics_django/web/format.py
@@ -105,6 +105,10 @@ def format_output(obj, expr, format=None):
             form_expr = Expression("StandardForm", expr)
             result = form_expr.format(obj, "System`StandardForm")
             return eval_boxes(result, result.boxes_to_svg, obj)
+        elif head in ("System`Symbol", "System`Integer"):
+            # What obviously doesn't need MathML? Note: System`Real is kept in MathML
+            # because we might have exponents, e.g. 3 * 10^8.
+            result = Expression("StandardForm", expr).format(obj, "System`OutputForm")
         else:
             result = Expression("StandardForm", expr).format(obj, "System`MathMLForm")
     else:


TiagoCavalcanteTrindade:
> For your consideration. Reducing the use of MathML where it isn't needed reduces the need for it to want to reformat it on loading.
>
> The output changes slightly, and there may be additional styling, or not.
>
> I am curious to know what you think and what your opinion is.

I want to remove MathML from everywhere it's possible, but the style changes here are too big (they're rendered like strings). It's hard to do the additional format as we need to check the value, we can think something was `True` but it was `"True"`.

Also, the re-renderings would stop with KaTex.

So I guess it's better to keep this surrounded by MathML by now.

rocky:
> > For your consideration. Reducing the use of MathML where it isn't needed reduces the need for it to want to reformat it on loading.
> > The output changes slightly, and there may be additional styling, or not.
> > I am curious to know what you think and what your opinion is.
> 
> I want to remove MathML from everywhere it's possible, but the style changes here are too big (they're rendered like strings). It's hard to do the additional format as we need to check the value, we can think something was `True` but it was `"True"`.

A couple things here. Inside format we know what the type is, so why is it so hard to let HTML know about this? 

And for strings those are surrounded by quotes. So the question there is why is the string style so different than MathML? 

> 
> Also, the re-renderings would stop with KaTex.

But wrong is still wrong and complicated is still complicated. If the cellhas nothing to do with formula typesetting why are we trying to fit it into a formula? 


> 
> So I guess it's better to keep this surrounded by MathML by now.

The conclusion I don't really care about as much as the reasoning behind it.

TiagoCavalcanteTrindade:
> A couple things here. Inside `format.py` we know what the type is, so why is it so hard to let HTML know about this?

@rocky the `results` part of the JSON is a string in the form of an XML element or text (thinking now the user shouldn't try to print a XML). We would or add a new field to the JSON or create a `FormulaWithoutMathMLForm`.

rocky:
> > A couple things here. Inside `format.py` we know what the type is, so why is it so hard to let HTML know about this?
> 
> @rocky the `results` part of the JSON is a string in the form of an XML element or text (thinking now the user shouldn't try to print a XML). 

Right ditch the XML. The structure there could be equally (if not more uniformly and simply) be accessed as JSON.


> We would or add a new field to the JSON 

I think putting all in JSON the best thing.


or create a `FormulaWithoutMathMLForm`.
> 
> > And for strings are surrounded by quotes. So the question there is why is the string style so different than MathML?
> 
> image

TiagoCavalcanteTrindade:
@rocky transform everything in JSON isn't the right thing, have you already tried to write a HTML page in JSON?

I was thinking about this and let MathML where it isn't necessary is wrong, I guess the best is wrap the booleans and integers in a HTML tag, I just don't know if it should be hardcoded or be in a `Form`.

rocky:
> 
> 
> @rocky transform everything in JSON isn't the right thing, have you already tried to write a HTML page in JSON?
> 
> When I said XML I was referring to MathML (which can be a form of XML).
> 
> I was thinking about this and let MathML where it isn't necessary is wrong, I guess the best is wrap the booleans and integers in a HTML tag,

Ok - use this as  starting point and continue. 

>  I just don't know if it should be hardcoded or be in a `Form`.

I don't recall ever suggesting anything like this.

PR #99, opened 2021-07-03
Misc bug and release testing

rocky:


mathics_django/web/format.py
```diff
@@ -43,8 +46,6 @@ def eval_boxes(result, fn, obj, **options):
     if isinstance(format, dict):
         return dict((k, obj.format_output(expr, f)) for k, f in format.items())
 
-    from mathics.core.expression import Expression, BoxError
```
TiagoCavalcanteTrindade, line 46: Why remove it from here? Isn't lazy reload good?

mathics_django/web/format.py
```diff
@@ -65,9 +66,7 @@ def eval_boxes(result, fn, obj, **options):
 
     # This part was derived from and the same as evaluation.py format_output.
 
-    use_quotes = (
-        Expression("Settings`$QuotedStrings").evaluate(obj).get_head().to_python()
-    )
+    use_quotes = get_settings_value(obj.definitions, "Settings`$QuotedStrings")
```
TiagoCavalcanteTrindade, line 69: Great!

mathics_django/web/format.py
rocky: You spend _way_ too much time worrying about this kind of stuff

Expression is needed to format just about everything. 

But tell you what - run some tests to see how much time you save doing this. I doubt you will find a difference. But prove me wrong!

TiagoCavalcanteTrindade: > You spend way too much time worrying about this kind of stuff

Ok, I'll try to stop doing this.

PR #104, opened 2021-07-11
Misc tweaks...

rocky:
version.py: bump to dev version
django_doc.py: Mathics core will be removing DocElement
web/format.py: comment in a debug print that is useful to uncomment.


TiagoCavalcanteTrindade (approved):
LGTM

PR #109, opened 2021-07-18
Misc tweaks...

rocky:
* `.gitignore`: ignore files derived from mathics-threejs-backend
* `django_doc.py`: variable rename: xml_data -> doc_data
* `docpipeline.py`: Mathics core no longer (and dangerously/inefficiently) initializes doc data when the module is loaded. So this needs to be done explicitly.


TiagoCavalcanteTrindade (approved):
LGTM

PR #110, opened 2021-07-18
More of the same

rocky:
@TiagoCavalcanteTrindade  with some of the recent changes on the Mathics doc side, some of showing previously-computed examples is broken. 

Fixing that (at least as well as it is now done on the Mathics core side) will be done in another PR. After this, I'd like to make sure the revision numbers get back into "about" 

I'll be doing a docker update with all of this today, after improving PyPI docker configs.


TiagoCavalcanteTrindade (commented):
LGTM

TiagoCavalcanteTrindade:
> After this, I'd like to make sure the revision numbers get back into "about"
Ok.

PR #111, opened 2021-07-18
Revise "about" page; mathics-threejs-backend again

rocky:
Reinstate mathics-threejs-backend code that seems to have been lost in various merges. 

But beef up `Makefile` to copy `mathics-three-js-backend`'s `package.json`. 

From `package.json` pick out revision numbers. 

Split up "about" page into different software sections.


.github/workflows/ubuntu.yml
```diff
@@ -20,15 +20,16 @@ jobs:
...
+        sudo apt-get update -qq && sudo apt-get install -qq liblapack-dev llvm-dev remake nodejs npm
         python -m pip install --upgrade pip
         # Can comment out when next Mathics is released
         python -m pip install -e git://github.com/mathics/Mathics#egg=Mathics3[full]
+        npm install
         pip install -e .[full]
     - name: Install Mathics Django
       run: |
-        make develop
+        remake -x develop
```
TiagoCavalcanteTrindade, line 31: Won't this just let the build time slower as we need to install a new dependency?

rocky: How much time do you estimate it takes. 

For my part as soon as something messes up all that time is lost in my mental ability to figure out what's wrong.

TiagoCavalcanteTrindade: It won't take so long. I didn't know it let the packaging easier.

rocky: Some of this _isn't_ hypothetical. If you look at the CI history we were failing in `make develop` and I needed to figure out why. And then I realized hey, I have spent _years_ writing a tool that will help on this. Why can't I just use it (even if it takes < 5 seconds more per run).

PR #116, opened 2021-07-19
Fix 3d graphics

TiagoCavalcanteTrindade:


rocky:
Looks much better. Here are the things I notice

The points no longer appear in: 
```
Graphics3D[Table[Point[{x, y, z}], {x, 5}, {y, 5}, {z, 5}]]
```

Axes in 

```
Plot3D[Sin[x y], {x, -2, 2}, {y, -2, 2}, Mesh->Full, PlotPoints->21, TicksStyle->{Darker[Magenta], Darker[Blue]}]
```

can now sometimes be obscured by the graph. Previously  axis labels swung around. This might be a threejs fix since I guess the old behavior might be been wrong.

The box in 

```
Graphics3D[{ Cylinder[{{1,1,1}, {10,10,10}}], Cylinder[{{-1,-1,-1}, {-10,-10,-10}}] }]
```

is much smaller than the other boxes. 

I am okay if you want with merging this in and then continuing to address other things as separate PRs though.

rocky (approved):

TiagoCavalcanteTrindade:
All of them are mathics-threejs-backend related, so I'll merge it now end launch a new mathics-threejs-backend later this week/next week.

TiagoCavalcanteTrindade:
>The points no longer appear in:
>
> ```
> Graphics3D[Table[Point[{x, y, z}], {x, 5}, {y, 5}, {z, 5}]]
> ```

Fixed [here](https://github.com/Mathics3/mathics-threejs-backend/commit/1acad7c908497870e2f31289c8c374c22be48828).

TiagoCavalcanteTrindade:
@rocky I discovered what is going on with small bounding boxes: the viewpoint is already scaled in Mathics Core and scaled here again, so the camera is farer than it should, and it's smaller than the others because the multiplier is bigger (as the primitives are bigger).

rocky:
> @rocky I discovered what is going on with small bounding boxes: the viewpoint is already scaled in Mathics Core and scaled here again, so the camera is farer than it should, and it's smaller than the others because the multiplier is bigger (as the primitives are bigger).

Good detective work! This is something that had been nagging me when you mentioned extent calculation: a nice way to divide the rendering to be done.  

In Mathematica this is "ImageSize" https://reference.wolfram.com/language/ref/ImageSize.html so if the value is "Automatic" then I guess Mathics has done this and if it is some other value, I guess it has done it as well.  Only if there is no ImageSize I guess the back end has total freedom. 

For now, probably we need to let Mathics core have its way. In the future though it would be nice to have a way to 

* tell mathics core to skip the ImageSize part for say json
* ensure that at the compat JSON what's up with ImageSize is explicit.

rocky:
Also it probably wouldn't be difficult in in the compact JSON to add unscaled values even when core has been scaled. That way the front end can ignore the calculation if it wants.

rocky:
> > The points no longer appear in:
> > ```
> > Graphics3D[Table[Point[{x, y, z}], {x, 5}, {y, 5}, {z, 5}]]
> > ```
> 
> Fixed [here](https://github.com/Mathics3/mathics-threejs-backend/commit/1acad7c908497870e2f31289c8c374c22be48828).

Just tried - works and is fast! too.

TiagoCavalcanteTrindade:
> The box in
> ```
> Graphics3D[{ Cylinder[{{1,1,1}, {10,10,10}}], Cylinder[{{-1,-1,-1}, {-10,-10,-10}}] }]
> ```
> is much smaller than the other boxes.

I've fixed it locally, now the problem that remains: the axes are not being shown correctly, also, what do you think of moving axes location calculation to mathics-threejs-backend (just add an option in Mathics Core that specifies the number of axes ticks)?

PR #118, opened 2021-07-21
More doc example fixups

rocky:
@TiagoCavalcanteTrindade while this still has problems, it is the closest to getting us back to where we were before I added another level of sectioning. 

I just don't understand how test output is handled. Actually a lot of this code feels a bit hacky. 

If you can figure out what's up and clean it up I'd be grateful.

On the other hand, I certainly don't _expect_ that. 

There will no doubt will be more cleanup iterations, including moving a lot of this code out of here and into its own repository.


rocky:
When I try this now things are really borked...

rocky:
For Omnibus testing, it is expedient to get  this into master now. There maybe a couple more changes along this line later.

PR #127, opened 2021-07-30
Add summary text in section lists

rocky:
There is a branch in Mathics core of the same name that starts to add the actual description text.


mathics_django/doc/django_doc.py
```diff
@@ -829,6 +836,13 @@ def __init__(
         self.section = section
         self.slug = slugify(title)
         self.title = title
+        if section:
```
rocky, line 839: This part is separate and an existing bug.

TiagoCavalcanteTrindade (approved):
Apart from the bug, it looks good.

PR #128, opened 2021-07-31
Reduce/DRY setting

rocky:
Reduce duplication with mathics.setting.
Reorganize settings file.


rocky:
Seems merged in already.

PR #142, opened 2021-10-27
Allow hiding startup message via URL query param

Techcable:
Hi I'm using Mathics inside a webview that doesn't support cookies/local storage.

Unfortunately, my local development environment is broken (for mathics),
so I haven't been able to test this (the javascript doesn't update).
However, this is a very simple change so I'm pretty sure it'll work (the javascript function 'shouldHideStartup()' does work in my simple smoke-tesets).

I'll work on my developer environment, but maybe you could consider accepting this anyways? I'd be happy to make a second PR to fix any bugs that showup.....

Also here's a link to the webview app (https://github.com/Techcable/mathicsd). I'm planning to do a public release soonish (it still needs a few minor tweaks :wink:).

### DIFF
--- a/mathics_django/web/media/js/mathics.js
+++ b/mathics_django/web/media/js/mathics.js
@@ -783,7 +783,17 @@ function domLoaded() {
 
 	MathJax.Hub.Configured();
 
-	if (localStorage.getItem('hideMathicsStartupMsg') === 'true') {
+	function shouldHideStartup() {
+		if (localStorage.getItem('hideMathicsStartupMsg') === 'true') {
+			return true;
+		}
+		let params = new URLSearchParams(document.location.search);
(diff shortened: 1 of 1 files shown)


TiagoCavalcante:
@rocky @mmatera this looks good to me, what do you think?

rocky:
As I said I think we should go the other way and add a GPL3 statement in the about. I don't see any advantage for people to make it easy to obscure the GPL3-ness.

Techcable:
@rocky seems pretty opposed to this, so I'm going to close this PR.

For my specific use case (mathicsd), I'm planning to redistribute a standalone binary. In that case, I can just apply the patch myself (this is fine because my project is already GPL licensed)

I respect your concerns with licensing. I was just hoping to make my life a little bit easier :smile:

PR #143, opened 2021-12-11
Migrate from django.conf.urls.url to django.urls.re_path

shirok1:
I ran mathicsserver and it show this:
```
File "***\site-packages\mathics_django\urls.py", line 5, in <module>
    from django.conf.urls import url, include
ImportError: cannot import name 'url' from 'django.conf.urls' (***\site-packages\django\conf\urls\__init__.py)
```
It turns out that django.conf.urls.url was removed in Django 4.0 . Replace them with django.urls.re_path seem to work on my PC.

### DIFF
--- a/mathics_django/urls.py
+++ b/mathics_django/urls.py
@@ -2,7 +2,7 @@
 # -*- coding: utf-8 -*-
 
 
-from django.conf.urls import url, include
+from django.urls import re_path, include
 from django.conf.urls.static import static
 from django.conf import settings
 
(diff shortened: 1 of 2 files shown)


rocky:
Thanks for noticing and improving. It looks like this works with _both_ django 3 _and_ django 4.

(At some point when we want _only_ django 4 and greater we will probably want to adjust https://github.com/Mathics3/mathics-django/blob/master/setup.py#L68)

PR #147, opened 2021-12-18
Link Copyright into Startup screens.

rocky:
This is recommended in https://www.gnu.org/licenses/gpl-3.0.html#howto

Unless there are objections this will go out in the next release probably on Sunday.


TiagoCavalcante (approved):
LGTM

PR #150, opened 2022-05-01
Update mathics-threejs-backend to 1.2.2

TiagoCavalcante:


rocky (approved):
LGTM - thanks for the package update to mathics-threejs-backend.

rocky:
@TiagoCavalcante Unfortunately this doesn't fix things.

It worked for a little while in chrome, but once it starts being broken it stays that way. Same in Firefox. 

Setting a breakpoint in `mathics_django/web/media/js/mathics.js`,  such as on line 140 or 115, these are never getting hit.

I see the the ajax calls with graphics3d in them but I don't think this is getting looked over as a dom element and I don't think drawgraphics3d is getting called. This would explain the behavior seen.

Inspecting one of the blank regions gives: 

```
<li class="result">
  <div style="display: block; width: 65vw; max-width: 400px; height: 65vw; max-height: 400px; padding-top: 5px; padding-bottom: 5px; position: relative; cursor: pointer; overflow: hidden; margin: auto;">
    <canvas style="display: block; width: 100%; height: 100%;" data-engine="three.js r138" width="0" height="0">
    </canvas>
</div>
</li>
```

TiagoCavalcante:
This is so weird, if it just didn't work, but it sometimes work and sometimes not. 
I'll investigate more it tomorrow.

PR #152, opened 2022-07-10
We need to use systemsymbols in formatting...

rocky:
Expression's first argument has to be a Symbol now.

Note that when we release we'll should bump the mathics core version to 5.0.0 since we have made incompatible changes.


TiagoCavalcante (approved):
LGTM

PR #153, opened 2022-07-11
Graphics3d fixup

rocky:
Go back to a version that doesn't use a custom shader.
    
    This is probably what has been preventing Graphics3D from working for the last
    several months.
    
    Error message:
    
    THREE.WebGLProgram: Shader Error 0 - VALIDATE_STATUS false
    
    Program Info Log: Must have a compiled vertex shader attached:
    SHADER_INFO_LOG:
    ERROR: 0:8: 'NaN' : undeclared identifier
    VERTEX
    
    ERROR: 0:8: 'NaN' : undeclared identifier
    
    1: #version 300 es
    2:                              in vec3 position;
    3:
    4:                              uniform mat4 projectionMatrix;
    5:                              uniform mat4 modelViewMatrix;
    6:
    7:                              void main() {
    8:                                      gl_PointSize = NaN;
    9:                                      gl_Position = projectionMatrix * modelViewMatrix * vec4(position, 1);
    10:                             }
    11:


rocky:
> @rocky I believe I can fix mathics-threejs-backend until Friday.

Great! - we won't merge this into master. But it is still useful to have around as a branch.

TiagoCavalcante:
@rocky I have released a new version of mathics-threejs-backend, this should be fixed now.

rocky:
@TiagoCavalcante When i try this in opera, I see that it works. However if I try it in Firefox or Chrome, I get the same thing. With Firefox, I tried starting up an entirely new browser. Recall that if things were working then I guess code can be cached so we don't see the problem.

PR #157, opened 2022-07-28
Allow text for URL links

rocky:
This is the Django part of https://github.com/Mathics3/mathics-core/pull/473

@TiagoCavalcante this reminds me that master is still broken for threejs. This weekend I plan on doing a general release. If we don't have something in place by then I'll go with graphics3d-fixup and we will have to deal with the custom shader version later.  It has now been months that this has been broken.  Whatever speedups or niceties there have been by using a custom shader seem to be way offset by the time and aggravation and breakage by this often not working.

### DIFF
--- a/mathics_django/doc/utils.py
+++ b/mathics_django/doc/utils.py
@@ -125,7 +125,13 @@ def escape_html(text, verbatim_mode=False, counters=None, single_line=False):
             if tag == "em":
                 return r"<em>%s</em>" % content
             elif tag == "url":
-                return r'<a href="%s">%s</a>' % (content, content)
+                try:
+                    text = match.group("text")
+                except IndexError:
+                    text = None
+                if text is None:
+                    text = content
+                return r'<a href="%s">%s</a>' % (content, text)
 
         text = HYPERTEXT_RE.sub(repl_hypertext, text)
 


TiagoCavalcante:
> @TiagoCavalcante this reminds me that master is still broken for threejs. This weekend I plan on doing a general release. If we don't have something in place by then I'll go with graphics3d-fixup and we will have to deal with the custom shader version later. It has now been months that this has been broken. Whatever speedups or niceties there have been by using a custom shader seem to be way offset by the time and aggravation and breakage by this often not working.

Well, I could call that a feature, and differently from what Microsoft does with bugs and security flaws in Windows, that would be more correct than wrong, because silently executing code with bugs isn't the best behavior.

What I mean is: mathics-core is passing JSON with nulls to mathics-threejs-backend, and in JS `null != undefined`, `undefined` is more-often-than-not the equivalent of Python `None`, and so only this is supported. That is why Symja and the [few] dozens of examples in the mathics-threejs-backend repo are working.

Unhappyly I won't have time to look at mathics-core this week.

TiagoCavalcante (approved):
LGTM

rocky:
> > @TiagoCavalcante this reminds me that master is still broken for threejs. This weekend I plan on doing a general release. If we don't have something in place by then I'll go with graphics3d-fixup and we will have to deal with the custom shader version later. It has now been months that this has been broken. Whatever speedups or niceties there have been by using a custom shader seem to be way offset by the time and aggravation and breakage by this often not working.
> 
> Well, I could call that a feature, and differently from what Microsoft does with bugs and security flaws in Windows, that would be more correct than wrong, because silently executing code with bugs isn't the best behavior.
> 
> What I mean is: mathics-core is passing JSON with nulls to mathics-threejs-backend, and in JS `null != undefined`, `undefined` is more-often-than-not the equivalent of Python `None`, and so only this is supported. That is why Symja and the [few] dozens of examples in the mathics-threejs-backend repo are working.

Ah. This is new and useful information. Thanks.

> 
> Unhappyly I won't have time to look at mathics-core this week.

Ok. What we'll do is put out a release. When this is fixed in core we can put out another core and django release. 

Moving to 5.0.0 (across most mathics packages) isn't supposed to be perfect. It is just to reflect what's gone on in a year, and there has been some breaking stuff.  If the change to mathics-core is just that we adhere to more correct JSON, then I don't see that as an API change. It might be a minor version change such as from 5.1.0, but I am not an expert on how semantic versioning is supposed to work.

PR #165, opened 2022-10-26
fixes OutputForm[...] and FullForm[...] results

mmatera:
This PR fixes the routine that formats the output, for the case of  `FullForm` and `OutputForm`. In master, 
```
OutputForm[f[x+b]]
```
produces
```
RowBox[List[(<string: "standardform"="">, <string: "["="">, RowBox[List[(<string: "fullform"="">, <string: "["="">, RowBox[List[(<string: "f"="">, <string: "["="">, RowBox[List[(<string: "plus"="">, <string: "["="">, RowBox[List[(<string: "b"="">, <string: ",="" "="">, <string: "x"="">)]], <string: "]"="">)]], <string: "]"="">)]], <string: "]"="">)]], <string: "]"="">)]]</string:></string:></string:></string:></string:></string:></string:></string:></string:></string:></string:></string:></string:></string:></string:>
```

With this fix, the string `f[x+b]` is shown.

### DIFF
--- a/mathics_django/web/format.py
+++ b/mathics_django/web/format.py
@@ -64,8 +64,8 @@ def format_output(obj, expr, format=None):
             expr = elements[0]
 
     if expr_type in ("System`FullForm", "System`OutputForm"):
-        result = Expression(SymbolStandardForm, expr).format(obj, expr_type)
-        return str(result)
+        result = expr.elements[0].format(obj, expr_type)
+        return result.boxes_to_text()
     elif expr_type == "System`Graphics":
         result = Expression(SymbolStandardForm, expr).format(obj, "System`MathMLForm")
 


mathics_django/web/format.py
```diff
@@ -64,8 +64,8 @@ def eval_boxes(result, fn, obj, **options):
             expr = elements[0]
 
     if expr_type in ("System`FullForm", "System`OutputForm"):
```
rocky, line 66: Although I wrote this code, we should now be used result.form somehow (which should be a symbol, not a name).
And testing on fonames is bad too.

In genera,l the output routines here and in mathiicsscript need revision and probably more changes to mathics-core.

rocky:
LGTM - I tried this in Django and see what is reported. See comment though.

mathics_django/web/format.py
mmatera: Yes, probably we should think about rewriting all of this when we finish the Makeboxes refactor.

mmatera: I was looking at this closer, and I see that the problem with using `result.form` is that this property is not available at the level of this function. The Result object is built in `mathics.core.evaluation.Evaluation.evaluate`, after calling `format_output`.

rocky: I have been thinking about this process. The current approach where somehow you declare at the beginning of the session what kind of format you want is wrong. 

Also, the front end should get back an expression, and make decisions based on what comes back what to do. 
So in this sense adding a "form" in result is less important since that can simply be done using `Head` of the expression. 

When issuing the top-level format, the front end should be able to declare which kinds of backends/renderers it can support, e.g. MathML, Asymptote, SVG, LaTeX, matplotlib, threejs, and so on. It can also give a priority to these backends to influence which ones it supports _better_ and this can help the formatter make choices when something can be rendered using one of several ways.

mmatera: > I have been thinking about this process. The current approach where somehow you declare at the beginning of the session what kind of format you want is wrong.
> 

WMA does something like that: the graphics front end uses `StandardForm`, while the command line interpreter uses `OutputForm` by default. Then, you can ask any other format, or modify the default by setting the `$Post` or the `$PrePrint` hooks. 

> As is done in Django, the front end gets back an expression, and from this it can then, based on what comes back, it can adjust parameters to the formatting.
> 
> So in this sense adding a "form" in result is less important since that can simply be done using `Head` of the expression.
> 


The idea of adding a "form" in the result was just a way to tag the output prompt without having access to the evaluation object. Here it seems that is not useful.

> When issuing the top-level format, the front end should be able to declare which kinds of backends/renderers it can support, e.g. MathML, Asymptote, SVG, LaTeX, matplotlib, threejs, and so on. It can also give a priority to these backends to influence which ones it supports _better_ and this can help the formatter make choices when something can be rendered using one of several ways.

I think that a good approach would be that the core returns a Boxed expression, and then the front-end could decide how to render it. However, for that, we should have a better implementation of the MakeBoxes/Formatter inside the core.

rocky: > that the core returns a Boxed expression,

What is the advantage of returning a Boxed expression as opposed to one that hasn't gone through that process?

A straight M-expression is much simpler to understand. Also, the format and capabilities may influence what gets boxed. 

For some forms `1 / 2` does not need numerator and denominator ever boxed, while for other forms like MathML or LaTeX, boxing numerator and denominator is useful before rendering. 

But from the front-end's perspective I don't see what it would do with the additional boxing, but rather it might just get in the way.

If one wants to provide a boxing without rendering step, okay, but I right now from the user's or front-end perspective I don't see right now any advantage.

mmatera: For the same reason that a web server should return html code, I guess.
A Box expression is an expression built on a subset of the WL that specified how a result should look like, and how it should be interpreted, and also that does not require an evaluation object to be rendered. But any other equivalent representation would work.

rocky: > For the same reason that a web server should return html code, I guess. A Box expression is an expression built on a subset of the WL that specified how a result should look like, and how it should be interpreted, and also that does not require an evaluation object to be rendered. But any other equivalent representation would work.

This sounds to me like you are making up analogies and chasing after whatever shiny new thing that comes along.

mmatera: Maybe it's that, or just the experience of building a homebrew message mechanism in the jupyter Wolfram kernel, and looking how the threejs backend works. But  OK, let's see how it should work when we reach that point.

PR #166, opened 2022-11-09
More ENV vars and fix HTTP {404,500}

rocky:
* Environment variables which change Django settings (also shown under Settings)
  - ``MATHICS_DJANGO_ALLOWED_HOSTS`` sets Django ``ALLOWED_HOST``; use semicolon to separate entries
  - ``MATHICS_DJANGO_DEBUG_HOSTS`` sets Django ``DEBUG``
  - ``MATHICS_DJANGO_DDISPLAY_EXCEPTIONS_HOSTS`` sets Django ``DISPLAY_EXCEPTIONS``

HTTP 404 and 500 rendering fixed. We had holdover code from Django 1.6!

### DIFF
--- a/mathics_django/web/templates/500.html
+++ b/mathics_django/web/templates/500.html
@@ -1,5 +1,6 @@
-{% extends "base.html" %}
-
 {% block content %}
-	<h1>Server error</h1>
+<h1>Server error</h1>
+<p>URI: {{request_path}}</p>
+<p>Sorry - no angry unicorn logos here. They are out to pasture.</p>
 {% endblock content %}
+{% extends "base.html" %}
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,16 @@
 CHANGES
 =======
 
+5.1.0.dev
+-----
+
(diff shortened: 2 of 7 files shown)


mathics_django/web/templates/500.html
```diff
@@ -1,5 +1,6 @@
...
-
 {% block content %}
-	<h1>Server error</h1>
-{% endblock content %}
\ No newline at end of file
+<h1>Server error</h1>
+<p>URI: {{request_path}}</p>
+<p>Sorry - no angry unicorn logos here. They are out to pasture.</p>
+{% endblock content %}
+{% extends "base.html" %}
```
TiagoCavalcante, line 6: Why this page extends base HTML but 404 don't?

TiagoCavalcante (approved):
LGTM - just a small thing

mathics_django/web/templates/500.html
rocky: When use that base, I get the useless logo header with the download, share, and load links, but without the images.

PR #168, opened 2022-12-10
Fix empty session_key on new sessions

### LINKED ISSUE #169: Session definitions are shared across all users
## Description

Every user accessing the Django website is sharing the same session (i.e. the same definitions, etc).

I think you can see this by looking at the cookies as well, because the users don't seem to have any session cookie set.

## How to Reproduce

1. Access the websites from two different isolated browsers tabs (from different devices on different networks as well)
2. In the first tab define something such as `f[x_] := x^2`
3. In the second tab try to use that definition with like `f[2]`

What I was expecting from the third step was the result `f[2]`, but what I'm getting is `4` even though the second isolated tab (or device) has never seen `f[x]` before.

## Your Environment

The versions of the modules are:
```
Mathics-Django==5.1.0.dev0
Mathics-Scanner==1.3.0.dev0
Mathics3==5.0.3.dev0
```
And they were all installed on Manjaro with `python3 setup.py install --user` from the main branch in the repos.

## Workarounds

By digging a bit I found that the `_evaluations` dictionary (which, as far as I understand, is used to store each user session) is being accessed by `session.session_key`, which seems to be always `None`.

I'm not familiar with Django, but after searching a bit I found that [a possible fix](https://stackoverflow.com/a/53585501) is to run `session.create()` on new sessions before using `session_key`, and that's what I implemented in #168.

When running `mathicsserver` with that change, the Django session cookie `sessionid` gets set as expected, which differentiates the session between each user.

alek3y:
The session was being shared by everyone because `session.session_key` was `None` when accessing the `_evaluations` dictionary.

From what I experimented this fixes #169.

### DIFF
--- a/mathics_django/web/models.py
+++ b/mathics_django/web/models.py
@@ -22,6 +22,7 @@ _evaluations = {}
 def get_session_evaluation(session):
     evaluation = _evaluations.get(session.session_key)
     if evaluation is None:
+        session.create()
         definitions = Definitions(add_builtin=True)
         # We set the formatter to "unformatted" so that we can use
         # our own custom formatter that understand better how to format


rocky (commented):
I think this is right, but I would like to understand more. 

Currently what are the specific problems you are encountering? Maybe open an issue to describe the environment and the kind of setup you have. What you see currently, and what you see with this fix. 

Thanks, and thanks for the patch. Sorry to take so long to respond.

alek3y:
Sure, my bad. I just opened #169.

rocky:
> Sure, my bad. I just opened #169.

Thanks for the additional information. Some of this may be murky with respect to when you want sessions to be shared or not. But let's try this and see how it goes.

PR #171, opened 2022-12-26
Split out controllers

rocky:
Controllers for documentation and about split off from monolithic "views.py".


rocky:
Although more could be done split out `views.py`, for example login kinds of things, or workspace manipulation kinds of things, documentation and the about are the two main things I care about.

PR #181, opened 2023-01-28
Reinstate PyMathics section in Documentation.

rocky:
We can't handle true guide sections yet though.

See pymathics.graph (branch 6.0.0) for a module that has largely been gone over.

### DIFF
--- a/mathics_django/doc/django_doc.py
+++ b/mathics_django/doc/django_doc.py
@@ -13,6 +13,7 @@ from types import ModuleType
 
 from django.utils.safestring import mark_safe
 from mathics import builtin, settings
+from mathics.core.pymathics import pymathics_builtins_by_module, pymathics_modules
 from mathics.core.util import IS_PYPY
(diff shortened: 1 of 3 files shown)


rocky:
Currently rendering pymathics.graph's Graph type doesn't work. Previously we did this via mathpotlib. With a newer threaded Django, this doesn't work because mathplotlib's standalone renderer needs to be run in the main thread.

In another PR though I will try to get inline SVG/PNG working and then we should be in even better shape than we were before.

TiagoCavalcante (approved):
LGTM

rocky:
Rendering SVG's from pymathics.graph now works! It wasn't such a big change once all of the support stuff was there. Basically we just needed to decode the matplotlib SVG image and strop off the initial `<xml>` tag. 

As always I think there will be other ways to do this. For example, pymathics.graph should eventually be changed to select whether to render via matplotlib or use Mathics3's graphics primitives.

PR #182, opened 2023-01-29
Better error handling

rocky:
Better handling and formatting of Syntax Errors and Python Exceptions.

views.py:
-------------
    
We now use the "form" parameter to detect syntax errors and to mark exception errors that are to be displayed.
    
This requires a recent Mathics 6.0.0
    
mathics.js:
---------------
    
Uses the tag set up for it in `views.py` to it to detect "Syntax Error" and "Python Exception" and formats these in a custom way that better displays these.
    
More work could be done for Python Exception in the area of custom traceback rendering. For example, file paths could be incorporated in link (<a>) tags.
    
Reformatted the file to remove tabbing and to make explicit (for Emacs) indent of 4 spaces.

### DIFF
--- a/mathics_django/web/media/js/mathics.js
+++ b/mathics_django/web/media/js/mathics.js
@@ -1,829 +1,862 @@
 // TODO: getSelection function
+// -*- mode: js; js-indent-level: 4; -*-
+
 
 let deleting,
-	blurredElement,
-	movedItem,
-	clickedQuery,
-	lastFocus = null,
-	welcome = true,
-	queryIndex = 0,
-	mouseDownEvent = null,
-	timeout = -1;
+    blurredElement,
+    movedItem,
+    clickedQuery,
+    lastFocus = null,
+    welcome = true,
+    queryIndex = 0,
+    mouseDownEvent = null,
+    timeout = -1;
 
 function getLetterWidth(element) {
-	const letter = document.createElement('span');
-	letter.innerText = 'm';
-	letter.style.fontFamily = element.style.fontFamily;
-	letter.style.fontSize = element.style.fontSize;
+    const letter = document.createElement('span');
+    letter.innerText = 'm';
+    letter.style.fontFamily = element.style.fontFamily;
+    letter.style.fontSize = element.style.fontSize;
 
-	document.body.appendChild(letter);
+    document.body.appendChild(letter);
 
-	const width = letter.getWidth();
+    const width = letter.getWidth();
 
-	document.body.removeChild(letter);
+    document.body.removeChild(letter);
 
-	return width;
+    return width;
 }
 
 function refreshInputSize(textarea) {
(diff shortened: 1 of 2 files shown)


rocky:
@mmatera and @TiagoCavalcante in order to get work I had planned for this weekend, this branch will go in as soon as it clears CI. There is another branch here that will inline SVG and PNG images which we will use for pymathics.graph.

As always, if you have comments or concerns not them, and it will be addressed in a future PR. Thanks.

mathics_django/web/views.py
```diff
@@ -293,18 +314,30 @@ def query(request: WSGIRequest) -> JsonResponse:
...
             raise
     result = {
         "results": [result.get_data() for result in results],
     }
     if settings.LOG_ON_CONSOLE:
+        from pprint import pprint as pp
+
         print(evaluation.timeout)
-        print(str(result))
+        pp(result)
```
rocky, line 340: Pretty-print output is just nicer here.

mathics_django/web/media/js/mathics.js
```diff
@@ -1,829 +1,881 @@
 // TODO: getSelection function
+// -*- mode: js; js-indent-level: 4; -*-
```
rocky, line 2: `.editorconfig`  in the root directory says indent_style is space and indent-level is 4. 
For some reason Emacs is not picking up editorconfig so this makes it explicit here.

rocky: Sorry for all of the space related changes below which turn tabs into spaces. I will mark the significant changes though.

mathics_django/web/media/js/mathics.js
```diff
@@ -1,829 +1,881 @@
...
+	bold.innerHTML = out.prefix + ":"
+	resultList.appendChild(bold);
 
-			resultList.appendChild(li);
-			resultList.style.display = 'block';
-		}
 
-		if (result.out.length) {
-			resultList.style.display = 'block';
+	// Syntax Error message
```
rocky, line 375: Here is where we start to handle syntax errors like `]` which produces a *Syntax::sntxb* error from the parser.

mathics_django/web/media/js/mathics.js
```diff
@@ -1,829 +1,881 @@
...
+	const pre = document.createElement('pre');
+	pre.innerHTML = out.text;
+
+	// Add this to a list element in a block (which is CSS styled with a nice
+	// frame around it.
+	resultList.appendChild(pre);
+	resultList.style.display = 'block';
+	li.appendChild(resultList);
+	list.appendChild(li);
+    } else if (results.length == 1 && results[0].form == "Python Exception") {
```
rocky, line 385: This is the other major tag detection change here where we do a rudimentary format of a traceback.

PR #183, opened 2023-01-29
More traceback tweaking...

rocky:
mathics.js: remove erroneous "," at beginning of lines due to .join();
            use .join("\n")
views.py: If we have a long traceback - over 19 lines, chop out the
            middle

Makefile: runserver-debug shows on console and shows traceback in web app


rocky:
Some of the many things I didn't address on the first round. Of course there is still room for improvement. Right now I am going for "passable".

PR #188, opened 2023-02-13
Add some pymathics.graph examples

rocky:
Extend gallery examples to include pymathics.graph


TiagoCavalcante (approved):
LGTM

mmatera:
What if `pymathics.graph` is not installed?

rocky:
You get the input you typed. But that is in my opinion Ok, because it alerts the user that there is more that is there. ;-)

PR #189, opened 2023-02-13
Makeboxes graph as standard

mmatera:
This branch removes the special support for formatting Graph, that was moved into the pymathics.graph module.


rocky:
LGTM

PR #190, opened 2023-02-16
remove spaces and line breaks inside urls

mmatera:
This single line PR is a proposal about how to handle long urls in docstrings. With this patch

```
<url>:NetworkX:
     https://networkx.org/documentation/networkx-2.8.8/reference/algorithms/\
      generated/networkx.algorithms.tree.mst.minimum_spanning_edges.html
</url>
```

the <url></url> tag is processed as
```
<a href=https://networkx.org/documentation/networkx-2.8.8/reference/algorithms/generated/networkx.algorithms.tree.mst.minimum_spanning_edges.html>NetworkX</a>
```

The pros are 
* we do not need to have long lines in documentation (flake8 complains about it)
* in the code, we can preserve the indentation, and visually looks better.
* we do not need to change the docstrings that were written like this.

The contra is that I don't know if sphinx supports something like that, so it would make the migration harder.

### DIFF
--- a/mathics_django/doc/utils.py
+++ b/mathics_django/doc/utils.py
@@ -124,12 +124,12 @@ def escape_html(text, verbatim_mode=False, counters=None, single_line=False):
             content = match.group("content")
             #
             # Sometimes it happens that the URL does not
-            # fit in 80 characters. Then, to avoid that 
+            # fit in 80 characters. Then, to avoid that
             # flake8 complains, and also to have a
             # nice and readable ASCII representation,
             # we would like to split the URL in several,
             # lines, having indentation spaces.
-            # 
+            #
             # The following line removes these extra
             # characters, which would spoil the URL,
             # producing a single line, space-free string.
--- a/mathics_django/doc/utils.py
+++ b/mathics_django/doc/utils.py
@@ -122,6 +122,18 @@ def escape_html(text, verbatim_mode=False, counters=None, single_line=False):
         def repl_hypertext(match):
             tag = match.group("tag")
             content = match.group("content")
+            #
+            # Sometimes it happens that the URL does not
+            # fit in 80 characters. Then, to avoid that 
+            # flake8 complains, and also to have a
+            # nice and readable ASCII representation,
+            # we would like to split the URL in several,
+            # lines, having indentation spaces.
+            # 
+            # The following line removes these extra
+            # characters, which would spoil the URL,
+            # producing a single line, space-free string.
+            #
             content = content.replace(" ", "").replace("\n", "")
             if tag == "em":
                 return r"<em>%s</em>" % content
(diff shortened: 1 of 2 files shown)


mathics_django/doc/utils.py
```diff
@@ -122,6 +122,7 @@ def repl_list(match):
         def repl_hypertext(match):
             tag = match.group("tag")
             content = match.group("content")
+            content = content.replace(" ", "").replace("\n", "")
```
rocky, line 125: Please put in a comment as to why this is being done.

rocky:
I am on the fence about this, but in an open-source project if this kind of thing makes you happy or removes an annoyance, and you want to do the work to do this _fully_ then sure, let's go ahead. 

So note that a change _also_ needs to be made in `doc_utils.py` of mathics-core. (In fact in looking over the generated LaTeX file is where I was first alerted something was amiss.)

And readthedocs for what is allowed in a URL tag is _also_ needed. 

> * we do not need to have long lines in documentation (flake8 complains about it)

flake8 has been told to lines up to 300 characters in mathics-core in `setup.cfg` so that can removed after URL handling allows spaces, if this is done.

> * in the code, we can preserve the indentation, and visually looks better.

yes, this is nice. 

> * we do not need to change the docstrings that were written like this.

This is kind of an unusual argument: I didn't understand how things worked, or test what I did, so instead I'll changed the code to match what thought or hoped it would be. 

Personally, I suppose I have done this many times myself too, and there is a kind of benefit:  this sometimes is how progress is made. However it is not something, personally, I advertise or tout as an advantage.

> The contra is that I don't know if sphinx supports something like that, so it would make the migration harder.

This is an interesting point too. Personally, if I didn't know if sphinx supported something, I would look. And if it doesn't I'd ask myself if not then _why_ not?  Or if it handles it but in a different way, then I might try to _match_ that rather than go invent a new convention on my own. 

Here is  situation with Sphinx: docs can live outside of Python code so it is not subject to pylint style. RsT reflows automatically, and you can use long lines. 

Redoing everything for Sphinx is indeed going to cause its own challenge. Not unlike the kind we have seen already. However, in my opinion having good docs _now_ is a sad conscious choice I've made. 

 Cutting over to Sphinx, even forgetting the markup is such a challenge and lower priority than all of those other major problems we have.

mathics_django/doc/utils.py
```diff
@@ -122,6 +122,19 @@ def repl_list(match):
         def repl_hypertext(match):
             tag = match.group("tag")
             content = match.group("content")
+            #
```
mmatera, line 125: @rocky , please check if this comment is OK.

rocky: @mmatera Excellent - thanks!

rocky:
@mmatera Looks like for this PR, I don't see anything more to do.

If you are up to changing mathics-core and readthedocs to bring them in line with this change. The please go ahead and merge this into master when you are satisfied. And then, at least in the short term (which who knows may be years), is s win.

PR #191, opened 2023-02-23
Add natlang examples, docpipeline etc.

rocky:
Makefile: use UTF-8 for XML doc generation
__init__.py make sure we import __version__
inout.js: Add NLP examples, comment out undirected edge use since that may have a problem using Combinatorica and pymathics.graph at the same time.


TiagoCavalcante (approved):
LGTM

PR #192, opened 2023-02-23
Reinstate the failing Graph examples...

rocky:
The problem is that combinatorica definitions and Mathics3 Graph define the same functions

### DIFF
--- a/mathics_django/web/media/js/inout.js
+++ b/mathics_django/web/media/js/inout.js
@@ -367,11 +367,10 @@ Graphics3D[Table[Point[{x, y, z}], {x, 5}, {y, 5}, {z, 5}]]',
 	'LoadModule["pymathics.graph"]',
 	'BinomialTree[3, DirectedEdges->True]',
 	'BalancedTree[3, 3]',
-	/*
-         There may be a a contention for UndirectedEdge here by combinatorica.
-         'g = Graph[{1 -> 2, 2 -> 3, 3 <-> 4}, VertexLabels->True]',
-	'ConnectedComponents[g]',
-        */
+	'(* Note that we need to specify context Pymathics` below because combinatorica also defines Graph, ConnectedComponents, and WeaklyConnectedComponents  *)',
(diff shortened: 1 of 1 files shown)


TiagoCavalcante (approved):
LGTM

mathics_django/web/media/js/inout.js
```diff
@@ -367,11 +367,10 @@ Graphics3D[Table[Point[{x, y, z}], {x, 5}, {y, 5}, {z, 5}]]',
...
 	'BalancedTree[3, 3]',
-	/*
-         There may be a a contention for UndirectedEdge here by combinatorica.
-         'g = Graph[{1 -> 2, 2 -> 3, 3 <-> 4}, VertexLabels->True]',
-	'ConnectedComponents[g]',
-        */
+	'(* Note that we need to specify context Pymathics` below because combinatorica also defines Graph, ConnectedComponents, and WeaklyConnectedComponents  *)',
+        'g = Pymathics`Graph[{1 -> 2, 2 -> 3, 3 <-> 4}, VertexLabels->True]',
+	'Pymathics`ConnectedComponents[g]',
+	'Pymathics`WeaklyConnectedComponents[g]',
```
mmatera, line 373: I wonder, the definitions are compatible?

rocky: > I wonder, the definitions are compatible?

Semantically I expect them to be same.  Especially since the Mathematica implementation started out with the Combinatorica definition.  At some point VertexNumber got renamed to VertexLabel. 

The basic problem is that what is in combinatorica and what is in the Mathics3 modules have two different interfaces and underlying implementations. In Combinatorica you need to wrap the graph in `ShowGraph[]` which converts the graph into boxes and lines.

And here is where we are different than Mathematica.  Because Mathematica is "curated" it would make a omnipotent decision to choose only one, possibly modified (which I imagine may upset contributors who have already published these things in a book and thus made the book obsolete.) 

For us, if we can get the context stuff correct, I see no reason why there can't be several interfaces.

PR #193, opened 2023-02-25
Update mathics-threejs-backend to 1.3.1 and stop using version without custom shaders

TiagoCavalcante:


rocky (commented):
LGTM - Many many Thanks!

PR #198, opened 2024-01-19
adding background and tooltips to Graphics3D

mmatera:
This PR restores the functionality of the `Background` option in Graphics3D, as well as provides the way to show a tooltip text in this kind of graphics object.

imagen


rocky:
Ok. Let's try it.

PR #199, opened 2024-01-19
Update to include new Polygon background

rocky:
* Revise Polygon background example to set background. 
* Untabify to make it clear what indentation is. 
* Split into multiple lines, long polygon example.

### DIFF
--- a/mathics_django/web/media/js/inout.js
+++ b/mathics_django/web/media/js/inout.js
@@ -303,7 +303,7 @@ function showGallery() {
             'TranslationTransform[{Sqrt[3] (i - j/2), 3 j/2}]' +
 	    '/@ {{Sqrt[3]/2, -1/2}, {0, 1}, {-Sqrt[3]/2, -1/2}}], ' +
             '{i, 7}, {j, i}]],' +
-            'Background->Orange]',
+            'Background->Red]',
 
         '(* Charts: *)',
         'BarChart[{{1, 2, 3}, {2, 3, 4}}]',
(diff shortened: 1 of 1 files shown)


mmatera:
@rocky, I do not understand the example. Some good examples to add would be

```
Graphics[Disk[], Background->RGBColor[.1,.3,1.]]
Graphics3D[Sphere[], Background->RGBColor[.1,.3,1.]]
```

and for errors,

```
Graphics[{Disk[],NoPrimitive[]}, Background->RGBColor[.1,.3,1.]]
Graphics3D[{Sphere[], NoPrimitive[]}, Background->RGBColor[.1,.3,1.]]
```

Notice that `Polygon` is not yet defined for 3D graphics, and that, with the last changes, the background is set as a true background, not by adding a primitive.

rocky:
Let's back up a little. 

The gallery is not a tutorial. It is not intended to _explain_ features, just show that they exist. Also we need to keep the _number_ of examples down small.

The documentation is there for understanding examples.  Something like  `Graphics[Disk[], Background->RGBColor[.1,.3,1.]]` is not very interesting not makes a catchy display. 

The example can be complicated if it makes a good visual. (Of course less complicated examples that make good visuals are even better. Here,  unfortunately we do not have the primitives that would simplify drawing this or other even better examples.)  

Therefore adding a background color to one of the _existing_  examples is the way to go.  I had also tried this on the example before but somehow to me it makes the bottom arrow of the star look funny. 

Possibly what you are suggesting should go in as an example in one for one of the examples in the documentation. Preferably as an example in a section where the examples are scarce. I am not seeing that we have "Background" documented anywhere. That is possibly where those examples might go. Another place is to  fill out RGBColor with more interesting examples and one that shows the use of real-valued R, G, and B values.

But this is done in mathics- core not here.

mmatera:
OK, in that case, LGTM

PR #201, opened 2024-02-01
compatibility with mathics-core PR 986

mmatera:
This PR is needed to keep compatibility with https://github.com/Mathics3/mathics-core/pull/986

PR #202, opened 2024-02-02
Address some doc bugs

rocky:
This tracks changes in https://github.com/Mathics3/mathics-core/pull/984

We need this in order to fix document display after loading in Mathics3 Modules.

### DIFF
--- a/mathics_django/doc/django_doc.py
+++ b/mathics_django/doc/django_doc.py
@@ -18,7 +18,7 @@ from mathics.doc.common_doc import (
     DocTests,
     DocText,
     Documentation,
-    XMLDoc,
+    DocumentationEntry,
     gather_tests,
     get_results_by_test,
     sorted_chapters,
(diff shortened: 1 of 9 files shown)


mathics_django/doc/django_doc.py
```diff
@@ -65,58 +70,27 @@ def get_title_html(self):
 
 
 class DjangoDocumentation(Documentation, DjangoDocElement):
```
mmatera, line 72: In the current mathics-core master branch, the `Documentation` class just has attributes related to the structure of the documentation, but not the part of loading `DocPart`s from modules, which is included in `MathicsMainDocumentation`. This class should subclass this other class.

mathics_django/doc/django_doc.py
```diff
@@ -11,12 +11,13 @@
...
-    DocGuideSection,
+    DocPart,
+    DocSection,
+    DocSubsection,
     DocTest,
     DocTests,
     DocText,
     Documentation,
-    Tests,
     XMLDoc,
```
mmatera, line 21: This class does not exist anymore. Now it was renamed `DocumentationEntry`

mmatera:
@rocky, I have implemented the minimal changes needed in #201. I will try to include the changes you made here in that other PR.

mathics_django/doc/django_doc.py
rocky: This has been chained in master now.  MathicsMainDocumentation was intended for testing only. LaTeX and HTML documentation are subclasses of Documentation.

mmatera:
@rocky, I guess this is deprecated, isn't it?

rocky:
yes, no longer valid.

PR #203, opened 2024-03-19
Improving guide handling

mmatera:
Changes needed to work together with https://github.com/Mathics3/mathics-core/pull/1025#issue-2194876664

### DIFF
--- a/.github/workflows/osx.yaml
+++ b/.github/workflows/osx.yaml
@@ -30,7 +30,7 @@ jobs:
         # For testing 3.11 we need to do something like the below until the next Mathics3 is released
         # python -m pip install -e git+https://github.com/Mathics3/mathics-scanner#egg=Mathics-Scanner[full]
         # python -m pip install -e git+https://github.com/Mathics3/mathics-core#egg=Mathics3[full]
-        git clone --branch main https://github.com/Mathics3/mathics-core
+        git clone --branch master https://github.com/Mathics3/mathics-core
         (cd mathics-core && pip3 install -e .[full])
         (cd mathics-core && bash ./admin-tools/make-op-tables.sh)
     - name: Install Mathics3 Django
--- a/.github/workflows/osx.yaml
+++ b/.github/workflows/osx.yaml
@@ -30,7 +30,7 @@ jobs:
         # For testing 3.11 we need to do something like the below until the next Mathics3 is released
         # python -m pip install -e git+https://github.com/Mathics3/mathics-scanner#egg=Mathics-Scanner[full]
         # python -m pip install -e git+https://github.com/Mathics3/mathics-core#egg=Mathics3[full]
-        git clone https://github.com/Mathics3/mathics-core
+        git clone --branch doc-code-rebased-rebased https://github.com/Mathics3/mathics-core
         (cd mathics-core && pip3 install -e .[full])
         (cd mathics-core && bash ./admin-tools/make-op-tables.sh)
     - name: Install Mathics3 Django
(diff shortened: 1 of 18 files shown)


mmatera:
This is how the documentation associated with a package looks:

imagen

imagen

And this is how it looks for a flat module:
imagen


Still, some aesthetic improvements must be done, like removing the blank lines in the index of a guide section.

mathics_django/web/controllers/doc.py
```diff
@@ -8,29 +8,52 @@
...
+seen_pymathics_modules = copy(pymathics_modules)
+
+MATHICS3_MODULES_SLUG = slugify(MATHICS3_MODULES_TITLE)
+
 
-def check_for_pymathics_load():
+def check_for_new_load_modules():
+    """
+    See if we have laoded any new Mathics3 modules since the last time
+    we checked. If so get an add the documenation for that.
```
rocky, line 34: documenation -> documentation

mathics_django/doc/django_doc.py
```diff
@@ -11,13 +11,14 @@
...
+    DocSection,
+    DocSubsection,
     DocTest,
     DocTests,
     DocText,
     Documentation,
-    Tests,
-    XMLDoc,
+    DocumentationEntry,
     gather_tests,
```
rocky, line 22: Something in my environment is telling me that `gather_tests` is no longer needed. 

Some thing for `import re` on line 8.

mathics_django/doc/django_doc.py
```diff
@@ -28,21 +29,25 @@
...
     doc_data = {}
 
 
-class DjangoDocElement(object):
+class DjangoDocElement:
+    """
+    Adds some HTML functions onto existing Django Document Elements.
+    """
+
     def href(self, ajax=False):
```
rocky, line 46: My environment tells me that something is weird with calling `self.get_url()`.  `get_url` is not defined in this class and it does not inherit from anything. Therefore were this to get called, we'd have an error.

rocky:
LGTM - Many good improvements and simplification of code. Thanks!  Merge when you are ready. See also the various comments should you want to address.

mathics_django/doc/django_doc.py
mmatera: I couldn't find `get_url` in this file. What is there is `get_uri`, which is defined and used to build the navigation links. `get_title_html` is defined for `DjangoDocElement`, and I think we need it for doing the documentation format.

.github/workflows/osx.yaml
```diff
@@ -30,7 +30,7 @@ jobs:
         # For testing 3.11 we need to do something like the below until the next Mathics3 is released
         # python -m pip install -e git+https://github.com/Mathics3/mathics-scanner#egg=Mathics-Scanner[full]
         # python -m pip install -e git+https://github.com/Mathics3/mathics-core#egg=Mathics3[full]
-        git clone --branch docpipeline_django_compat https://github.com/Mathics3/mathics-core
+        git clone --branch main https://github.com/Mathics3/mathics-core
```
rocky, line 33: Of course, should be "master", not "main".

mathics_django/doc/django_doc.py
rocky: What the linter is saying is that _if_ one tries to use  it, you'll get an error. 

I did some git archeology and this was added by me around 2020 when I split this off from what was in `mathics-core`.  There was a _lot_ of guess work there. 

There is no such thing as "building" documentation for Django. So we can remove this.

.github/workflows/osx.yaml
mmatera: Indeed...

mathics_django/doc/django_doc.py
mmatera: OK, but both functions seem to work properly, as they are called when navigating documentation. I added some comments in the docstrings to clarify where they are used.

PR #216, opened 2025-01-10
Import autoload_files from mathics.session

mmatera:


mmatera:
Close because now this is trivial.

PR #221, opened 2025-01-26
Improve support for LaTeX via MathML in documentation

mmatera:
This PR contains the minimal changes needed to support LaTeX expressions in the documentation.


mmatera:
imagen

rocky:
LGTM -  thanks. 

This starts us toward improving the online doc so it has can have an impact larger than what is visible right now. I suspect our homegrown parsing limits what we can offer MathML. For example, right now we have to tag `$x$^2` rather than `$x^2$`. 

Some things to note: 

While this works in a "dd" tag, I don't think it works in "dt".  I tried putting `$\sigma_k$`  in the `<dt>` portion of `DivisorSigma` , and that came out untranslated.

Also, when I look at `Cuboid`, is don't see `p_min`  (in the dd portion since dt doesn't work) getting rendered with a subscript. Know why?

At any rate, these can be done in a future PR.

rocky:
> OK, but we use `<dt>` blocks for code, so we do not need LaTeX code there.

That is not the way "dt" is supposed to work. A lot of our formatting and WMA's formatting use Mathematical notation in the definition term.

> 
> > Also, when I look at `Cuboid`, is don't see `p_min` (in the dd portion since dt doesn't work) getting rendered with a subscript. Know why?
> 
> Maybe because `p_min` is not wrapped in `$` ?

No. See https://github.com/Mathics3/mathics-core/blob/master/mathics/builtin/drawing/graphics3d.py#L257

mmatera:
Yep. The problem is a concatenation of several small issues:
* This regular expression
   `LATEX_RE = re.compile(r"(\s?)\$(\w+?)\$(\s?)")` defined in [mathics.doc.doc_entries](https://github.com/Mathics3/mathics-core/blob/275a3b8da9e6c2edfb7789d1ed5dbab12ef066ba/mathics/doc/doc_entries.py#L69C1-L69C47 ) is used in https://github.com/Mathics3/mathics-django/blob/2ed4a3d17b4f1485f94e35f0209d7938697f3bc6/mathics_django/doc/utils.py#L80  to replace ` $var$ ` by `<VAR>var</VAR> `. This makes sense if `var` is a sequence of alphabetic characters and numbers, but not for a LaTeX expression.  To avoid using it for more complicated expressions, we should change the regular expression to
`LATEX_RE = re.compile(r"(\s?)\$([A-Za-z]+?)\$(\s?)")`

* Due to the quotes, 'Cuboid[$p_min$]'  is translated to HTML as `<code>Cuboid[$p_min$]</code>`, which makes the code to be handled as "code" i.e., verbatim. The way to allow MathML in the arguments would be to use the quotes around the HEAD of the expression, like this: `'Cuboid'[$p_{min}$]`
* LaTeX interpreter understand `p_min` as `p_{m}in`, i.e., just the first character is considered as the subindex. To have `min` as subindex, we need to wrap it in curly brackets.

PR #223, opened 2025-01-27
Release 8.0.0

rocky:
Release 8.0.0

* Adjust CI to use wheels
* Update CHANGES
* Update version and package dependencies


rocky:
There is something fishy about having to run the tests from git. Possibly tables are not getting installed in mathics-core properly. However, fixing this is probably on the mathics-core side and probably will require another release.

Will do that though after this wave of releases is done.

Issue #228, opened 2025-02-13
Service /media/ urls from a relative not absolute URL


adamboutcher:
The media files are all defined using absolute URL and not relative which causes issues if trying to service the app through a proxy.

adamboutcher:
So I understand the need for the scripts to use absolute URLs so they serve correctly if you're using subdirs for fuctions etc

I've had to hack to get around this with

`# URL that handles the media served from STATIC_ROOT.
static_url = os.environ.get("MATHICS_DJANGO_URL", None)
if static_url is not None:
    STATIC_URL = static_url+"/media/"
else:
    STATIC_URL = "/media/"
`

rocky:
Ok  - please work this up into a PR.

A shorter way to write the last 4 lines is: 

```python
STATIC_URL = status_url+"/media" if static_url is None else "/media"
```

rocky:
> > Ok - please work this up into a PR.
> > A shorter way to write the last 4 lines is:
> > STATIC_URL = status_url+"/media" if static_url is None else "/media"
> 
> My original post didnt reliably work I have however edited this.

Ok. Good to hear. If you want to get the change into the code base, please put in a PR when you can.

You are in the best position to test whether things work. What I can do is verify that it doesn't break the behavior I use.

PR #230, opened 2025-03-08
Proxy support

adamboutcher:
An additional (see https://github.com/Mathics3/mathics-django/pull/229) change to fix documentation not working.


adamboutcher:
Related to https://github.com/Mathics3/mathics-django/issues/228

PR #250, opened 2026-04-01
Misc fixes and tweaks

rocky:
views.py: reset iteration count after each cell evaluation Makefile: "make runserver" should no set LOG_ON_CONSOLE, use make runserver-debug for that
views.py: Detect `$Aborted` and `$Failed` sooner
inout.js: Show example where we set PrintPrecision to 6, like in a Notebook

### DIFF
--- a/mathics_django/web/models.py
+++ b/mathics_django/web/models.py
@@ -37,7 +37,6 @@ def get_session_evaluation(session):
             evaluation, expr, format
         )
         autoload_files(definitions, ROOT_DIR, "autoload")
-    evaluation.iteration_count = 0
     return evaluation
 
 
--- a/mathics_django/web/models.py
+++ b/mathics_django/web/models.py
@@ -37,6 +37,7 @@ def get_session_evaluation(session):
             evaluation, expr, format
         )
         autoload_files(definitions, ROOT_DIR, "autoload")
+    evaluation.iteration_count = 0
     return evaluation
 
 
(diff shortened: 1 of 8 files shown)


mathics_django/web/views.py
```diff
@@ -267,6 +267,7 @@ def query(request: WSGIRequest) -> JsonResponse:
         query_log.save()
 
     evaluation = get_session_evaluation(request.session)
+    evaluation.iteration_count = 0
```
rocky: @mmatera We didn't need to do this before. I am sure why it is needed now or whether this is the best way to do.

Your thoughts?

mathics_django/web/models.py
```diff
@@ -37,6 +37,7 @@ def get_session_evaluation(session):
             evaluation, expr, format
         )
         autoload_files(definitions, ROOT_DIR, "autoload")
+    evaluation.iteration_count = 0
```
rocky: @mmatera We didn't need to do this before. I am sure why it is needed now or whether this is the best way to do.

Your thoughts?

mmatera: <strike>It shouldn't. MathicsSession set this to 0 on each evaluation...</strike>
`Evaluation.evaluation` should set this to 0, but it doesn't.

rocky: Thanks! I'll put in a separate PR for this.

PR #253, opened 2026-04-10
Add Settings`RenderTeXForm

rocky:
If True (the default) TeXForm output is rendered using MathJax. Otherwise it appears as LaTeX output.


mmatera (commented):
LGTM

